PR #1, opened 2022-03-08
Rename env.example to .env.example

ikerexxe:


pbrezina:
My opinion is that example files should not be hidden so they are visible. What is the intention to hide it? Is it only to have the same name as the `.env` file?

pbrezina:
If yes, I would rather prefer to fix the readme file to `cp env.example .env` instead.

ikerexxe:
I was trying to be consistent with the documentation, where there are several references to `.env.example`. The proposed change was easier to perform, but I don't mind reverting this change and changing the documentation instead (`.env.example` -> `env.example`).

pbrezina:
Please, update the documentation instead. I believe it is better to leave example file visible.

pbrezina (approved):

PR #12, opened 2022-09-22
Fix containers .ssh/authorized_keys

jakub-vavra-cz:
The authorized keys contained private instead of public key breaking the ssh authentication to the containers.


pbrezina (approved):
Thank you.

PR #15, opened 2022-10-18
containers: add nfs server

pbrezina:
This PR adds nfs server in separate container that can be used to test autofs.


src/ansible/roles/nfs/tasks/main.yml
```diff
@@ -0,0 +1,18 @@
...
+    name: '{{ item }}.service'
+    enabled: yes
+    state: started
+  with_items:
+  - nfs-server
+  - rpcbind
+
+# Due to limitations of container file system /exports must be mounted as a
+# volume and all shared directories must be created under /exports. This
+# is to make sure that the folder always exist.
```
ikerexxe, line 12: s/exist/exists

src/docker-compose.build.yml
```diff
@@ -14,3 +14,6 @@ services:
   client:
     image: localhost/sssd/ci-base-client:${TAG}
     container_name: sssd-wip-client
+  nfs:
+    image: localhost/sssd/ci-base-nfs:${TAG}
+    container_name: sssd-wip-nfs
```
ikerexxe, line 19: Missing EOF

data/ssh-keys/hosts/nfs.ecdsa_key
```diff
@@ -0,0 +1,9 @@
+-----BEGIN OPENSSH PRIVATE KEY-----
```
ikerexxe, line 1: I guess that adding the private key is intentional, right?

ikerexxe (commented):
I added some minor comments. On top of that, can you update the [flowchart](https://github.com/SSSD/sssd-ci-containers/blame/master/readme.md#L272) with the new image?

data/ssh-keys/hosts/nfs.ecdsa_key
pbrezina: Yes.

src/docker-compose.build.yml
pbrezina: Fixed.

pbrezina:
Done. I also workaround the need for /exports directory so I removed it. (Tests now use /dev/shm/exports which is always tmpfs in the containers so can hold nfs shares).

ikerexxe (approved):
LGTM. Thanks!

PR #18, opened 2022-12-02
containers: add kdc

pbrezina:
Adding a standalone KDC container.


src/ansible/roles/packages/tasks/Fedora.yml
```diff
@@ -111,6 +111,17 @@
       - nfs-utils
   when: inventory_hostname == 'base-nfs'
 
+- name: Install packages for KDC base image
```
ikerexxe, line 114: Do we need to install the packages in CentOS images?

ikerexxe (commented):
Just one question inline.

Apart from that you should take a look at the CI failures.

src/ansible/roles/packages/tasks/Fedora.yml
pbrezina: CentOS include Fedora.yml, so it gets installed there.

PR #21, opened 2023-02-02
provision: make playbooks work on virtual machines in idm-ci

jakub-vavra-cz:
Create playbook_vm.yml to include image_base and image_service to work on VM with become.
Modify packages role so it recognizes rhel and applies pieces on appropriate host groups.
Replace freeipa with ipa for RHEL.
Improve ldap and ipa task idempotency.


src/ansible/roles/ad/tasks/main.yml
```diff
@@ -1,3 +1,17 @@
+- name: 'Create netbios name for a machine in cloud'
+  set_fact:
+    netbios: "AD{{ lookup('password', '/tmp/ad_netbios chars=ascii_uppercase,digits length=4', seed=ansible_host) }}"
+  when: 'ansible_host != "172.16.200.10"'
```
pbrezina, line 4: Relying on IP address is not very nice. Wouldn't it be better to introduce a variable like `CONTAINERS=yes/no` or `TARGET_ENVIRONMENT=containers|idm-ci|...possible-future-envs`?

pbrezina: Also instead of settings facts conditionally here (and possibly other places), wouldn't be better to put it to idm-ci specific variable file?

src/ansible/roles/common/tasks/main.yml
```diff
@@ -12,9 +12,15 @@
...
+- name: Create /data
+  file:
+    path: '/data'
+    state: directory
+    mode: 0700
+
 - name: Copy common data
   synchronize:
-    src: '{{ playbook_dir }}/../../data/'
+    src: '{{ role_path }}/../../../../data/'
```
pbrezina, line 23: Path to data directory can also go to variable file.

src/ansible/roles/ipa/tasks/main.yml
```diff
@@ -39,27 +39,39 @@
     ipa automountmap-del default auto.direct
   args:
     stdin: '{{ service.ipa.password }}'
+  register: automountmapdel
+  failed_when:
+    - 'automountmapdel.rc != 0 and "automount map not found" not in automountmapdel.stderr'
```
pbrezina, line 44: Why?

pbrezina (commented):
I didn't read it all yet, but here are first comments.

src/ansible/roles/ipa/tasks/main.yml
jakub-vavra-cz: When running second time it is already deleted so "not found" is actually OK state.

src/ansible/roles/ad/tasks/main.yml
jakub-vavra-cz: This specific address is hardcoded all around this playbook so relaying on it does not seem so bad.
Having variable and variable file for a specific environments sounds like a good idea.

As for the this AD playbook: I was hoping that it would be easy to make it work so I can get the framework running properly in idmci on VMs, and enable others to work on this. The end goal is to use the idm-ci playbooks that have support for ad forests instead.

src/ansible/roles/ipa/tasks/main.yml
jakub-vavra-cz: There were several other places where the playbook failed on second run because something was already done.

src/ansible/roles/common/tasks/main.yml
jakub-vavra-cz: I had to touch this as I was running the playbook from other ansible and the plabook_dir was not where it expected. Having variable for data dir path would definitely make it cleaner.

jakub-vavra-cz: Done.

src/ansible/roles/ad/tasks/main.yml
jakub-vavra-cz: AD playbook is not used after all.

src/ansible/roles/common/tasks/main.yml
```diff
@@ -1,3 +1,7 @@
+- name: Set data path
+  set_fact:
+    data_path: '{{ role_path }}/../../../../data'
```
pbrezina, line 3: Why is this needed? The playbooks work perfectly with playbook_dir.

src/ansible/roles/dns/tasks/main.yml
```diff
@@ -0,0 +1,30 @@
+- name: Install dnsmasq package
```
pbrezina, line 1: Mrack should get new dnsmasq output to generate dnsmasq config and this should go into shared playbooks so other projects can benefit from it as well.

src/ansible/roles/packages/tasks/CentOS8.yml
```diff
@@ -30,7 +30,7 @@
   - name: Enable IdM module
     shell: |
       dnf module enable -y idm:DL1
-  when: "'base_ground' in group_names"
+  when: "'base_ground' in group_names or 'base' in group_names"
```
pbrezina, line 33: This won't work for containers as the step will be repeated every time a new layer is added.

src/ansible/roles/packages/tasks/RedHat.yml
```diff
@@ -1,6 +1,173 @@
-- name: 'Packages are the same as in CentOS {{ ansible_distribution_major_version }}'
-  include_tasks: '{{ include_centos }}'
-  loop_control:
-    loop_var: include_centos
-  with_first_found:
-  - files: '{{ "CentOS" | distro_includes(ansible_distribution_major_version) }}'
+- name: Install packages for ground base image
```
pbrezina, line 1: This change is not needed. Packages can be perfectly shared with CentOS.

src/ansible/variables.yml
```diff
@@ -45,7 +45,7 @@ user: {
     password: Secret123
   },
   regular: {
-    uid: 1000,
+    uid: 1024,
```
pbrezina, line 48: This is not needed. The variable should be overwritten when running the playbook from te metadata.

pbrezina (changes requested):

src/ansible/playbook_vm.yml
```diff
@@ -0,0 +1,68 @@
+---
+# This playbook is to apply tasks similar to image_base and image_service
+# on virtual machines instead of containers as the roles need become true there
+# Some of the pieces also do not make any sense to apply on VMs.
```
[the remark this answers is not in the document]
jakub-vavra-cz: Become true does not not work for containers, it crashes on lack of sudo (Been there, seen that, the code that worked around that was plain ugly).
There is no use for cleaning up package manager caches on the VMs as it just delays the whole process and You need to rebuild/download them again for installing other pieces.

The ssh_server task just did work with VMs as they had different inventory_hostname and it just broken the setup done by idm-ci.

src/ansible/roles/dns/tasks/main.yml
jakub-vavra-cz: Well I thought it would make a sense to have the dns the same rol so it can be used also for container with a bit of tuning.

src/ansible/variables.yml
jakub-vavra-cz: It does need to be overridden this way so it is clearly better this way.

src/ansible/roles/packages/tasks/RedHat.yml
jakub-vavra-cz: You clearly did not spot that for RedHat I am installing ipa that is used for testing in our tests instead of freeipa. There are also some debuginfo packages not available on RHEL so I am ignoring them for now (so the lack is visible and can be revisited later).

src/ansible/roles/packages/tasks/CentOS8.yml
jakub-vavra-cz: Ok, I probably misundestood that when we wer btalking about that.
I will fix it.

jakub-vavra-cz: Fixed.

src/ansible/roles/dns/tasks/main.yml
jakub-vavra-cz: We can do that in future but for now it make sense to have this here instead of having the dns setup copied in every metadata.

jakub-vavra-cz:
@pbrezina  : Please re-review.

pbrezina:
Thank you, I will look at it tomorrow.

src/ansible/group_vars/all
```diff
@@ -56,3 +56,13 @@ user: {
     password: Secret123
   }
 }
+
+free_ipa_packages: {
```
pbrezina, line 60: I don't really care, but AFAIK the brand is "freeipa" and it should not be split.

src/ansible/roles/dns/templates/etc.dnsmasq.conf.j2
```diff
@@ -0,0 +1,81 @@
+# This config is specific to our use and does not reflect multiple AD DCs
+# and other setups.
+# TODO: Add all AD servers
+# TODO: Remove meta_ip dependency (from  idm-ci).
+# TODO: Generalize this for all machines.
+# dnsmaq config
```
pbrezina, line 6: typo: dnsmasq

src/ansible/roles/dns/templates/etc.dnsmasq.conf.j2
```diff
@@ -0,0 +1,81 @@
+# This config is specific to our use and does not reflect multiple AD DCs
```
pbrezina, line 1: This config is specific to downstream CI use case

src/ansible/roles/dns/templates/etc.dnsmasq.conf.j2
```diff
@@ -0,0 +1,81 @@
...
+{% endif %}
+{% if 'kdc.test' in hostvars %}
+ptr-record={{ hostvars['kdc.test']['meta_ip'].split('.') | reverse | join(".") }}.in-addr.arpa,kdc.test
+{% endif %}
+{% if 'nfs.test' in hostvars %}
+ptr-record={{ hostvars['nfs.test']['meta_ip'].split('.') | reverse | join(".") }}.in-addr.arpa,nfs.test
+{% endif %}
+
+
+#hostvars['nfs.test']['meta_ip']
```
pbrezina, line 81: Remove?

src/ansible/roles/ipa/tasks/main.yml
pbrezina, line 44: Please, deindent to follow coding style that is used for these playbooks.

src/ansible/roles/dns/tasks/main.yml
```diff
@@ -0,0 +1,30 @@
...
+  ansible.builtin.package:
```
pbrezina, line 2: Please, avoid ansible.builtin prefix to follow coding style from other files.

src/ansible/roles/packages/tasks/Fedora.yml
```diff
@@ -1,38 +1,32 @@
...
+    package:
       state: present
       name:
       - dnf-plugins-core
-
-  - name: Install additional repositories
-    shell: |
-      dnf config-manager --add-repo {{ item }}
-    with_items:
-    - https://cli.github.com/packages/rpm/gh-cli.repo
```
pbrezina, line 13: We want this on every container, not just the development one.

src/ansible/roles/packages/tasks/Fedora.yml
```diff
@@ -45,95 +39,103 @@
...
       - systemd
       - tcpdump
-      - tig
-      - tmate
-      - tmux
-      - vim
       - wget
       - which
       - wireshark-cli
+      disable_gpg_check: yes
```
pbrezina, line 45: Why? This definitely should not be needed.

src/ansible/roles/packages/tasks/Fedora.yml
```diff
@@ -45,95 +39,103 @@
       - sudo
       - systemd
       - tcpdump
-      - tig
-      - tmate
-      - tmux
-      - vim
```
pbrezina, line 51: We want these on every container, not just the devel one. Please, avoid functional changes.

src/ansible/roles/packages/tasks/Fedora.yml
```diff
@@ -45,95 +39,103 @@
...
-        glibc                   \
-        libcmocka               \
-        libdhash                \
-        libini_config           \
-        libldb                  \
-        libtalloc               \
-        libtevent
-    when: debuginfo
-  when: "'base_client' in group_names"
+    command: dnf debuginfo-install  -y  {{ item }}
```
pbrezina, line 63: Why this change?

src/ansible/roles/packages/tasks/Fedora.yml
```diff
@@ -45,95 +39,103 @@
...
+      disable_gpg_check: yes
   when: "'base_ground' in group_names"
 
 - name: Install packages for client base image
   block:
+  - name: Install ipa client
+    package:
+      state: present
+      name: '{{ ipa.client }}'
   - name: Install SSSD and its dependencies
```
pbrezina, line 54: Separate with blank line.

src/ansible/roles/packages/tasks/Fedora.yml
```diff
@@ -45,95 +39,103 @@
...
+      - libdhash
+      - libini_config
+      - libldb
+      - libtalloc
+      - libtevent
+    ignore_errors: yes
+    when:
+     - debuginfo
+     - "'dnf' in ansible_facts.packages"
+  - name: Install packages required for passkey testing
```
pbrezina, line 77: This is handled on other place.

src/ansible/roles/packages/tasks/Fedora.yml
```diff
@@ -45,95 +39,103 @@
...
+    package:
+      state: present
+      name:
+      - gcc
+      - openssl-devel
+      - umockdev
+    when: passkey_support
+  when: "'base_client' in group_names or 'client' in group_names or 'base_ipa' in group_names or 'ipa' in group_names"
```
pbrezina, line 85: Again, avoid functional changes. This block is for client only.

src/ansible/roles/common/tasks/main.yml
```diff
@@ -65,9 +65,19 @@
     mode: '0600'
   with_items:
   - { src: 'root.id_rsa', dest: 'id_rsa' }
-  - { src: 'root.id_rsa.pub', dest: 'authorized_keys' }
+#  - { src: 'root.id_rsa.pub', dest: 'authorized_keys' }
   - { src: 'root.id_rsa.pub', dest: 'id_rsa.pub' }
 
+- name: Append or create authorized_keys
```
pbrezina, line 71: Ok, but why?

src/ansible/roles/ssh_server/tasks/main.yml
```diff
@@ -11,3 +11,7 @@
     name: '{{ systemd.services.sshd }}'
     enabled: yes
     state: restarted
+  register: restart_res
```
pbrezina, line 14: What issues did you see? I never run into any issue when provisioning this in downstream ci.

src/ansible/playbook_vm.yml
```diff
@@ -8,11 +8,13 @@
 - name: Include base
   ansible.builtin.import_playbook: playbook_image_base.yml
   vars:
+    passkey_support: no
```
pbrezina, line 11: How are you going to test passkey on RHEL?

pbrezina (changes requested):
Hi Jakub, please, see some comments inside. And:

* Do not introduce any functional or breaking changes, this repository is for containers and upstream use case first and things are the way they are for a reason. Especially since neither change you have here is required for provisioning virtual machines, it works as it is already.
* Red Hat must provide exactly the same content as other distributions. If you want to remove some things out to provide minimal testing environment, please do so with a switch that can be applied on all distributions so you have the same setup regardless of distribution.
* The whole conversion from dnf to package is pointless. RHEL runs DNF, dnf package is installed on RHEL and yum is just symlink to dnf. Please avoid this change.
* We don't have use for idempotency since you are not going to provision machine that was already setup, it just complicates stuff. For testing, you can just start from specific task with `--start-at-task `. If you insist on idempotency, then please, implement it on all tasks.

src/ansible/roles/dns/templates/etc.dnsmasq.conf.j2
jakub-vavra-cz: removed.

src/ansible/roles/dns/templates/etc.dnsmasq.conf.j2
jakub-vavra-cz: Rephrased.

src/ansible/group_vars/all
jakub-vavra-cz: changed

src/ansible/playbook_vm.yml
jakub-vavra-cz: Well as we discussed before, we do not want to pull the whole build environment that is needed for passkey by default. There was discussion about packaging the paskey pieces in copr an pulling them when needed instead.

src/ansible/roles/common/tasks/main.yml
jakub-vavra-cz: Because this code just overwrote the authorized keys ignoring that there are already authorized_keys from idmci. This way it works nicely together instead of breaking the idmci tooling.

src/ansible/roles/ssh_server/tasks/main.yml
jakub-vavra-cz: I had a multiple completely random failures there probably due to openstack being under bigger load.

src/ansible/roles/ipa/tasks/main.yml
jakub-vavra-cz: Done.

src/ansible/roles/packages/tasks/Fedora.yml
jakub-vavra-cz: Well, we do not want these on anything else, they are not needed nor wanted for test environment. 
None of these is needed for testing and some of them are missing from rhel baseos/appstream

src/ansible/roles/packages/tasks/Fedora.yml
jakub-vavra-cz: This is needed for internal repos.

src/ansible/roles/packages/tasks/Fedora.yml
jakub-vavra-cz: done

src/ansible/roles/packages/tasks/Fedora.yml
jakub-vavra-cz: Because some of these are not in baseos/appstream and we do not want to taint the machines with pulling things from buildroot.

This way whatever is available is installed and the whole transaction does not fail.

Consider this a direct result of the given constraint to use/include Fedora.yml instead of having a separate standalone RedHat.yaml

src/ansible/roles/dns/tasks/main.yml
jakub-vavra-cz: Ack

src/ansible/roles/packages/tasks/Fedora.yml
jakub-vavra-cz: AFAIK it was there when I was rebasing.

jakub-vavra-cz:
> Hi Jakub, please, see some comments inside. And:
> 
>     * Do not introduce any functional or breaking changes, this repository is for containers and upstream use case first and things are the way they are for a reason. Especially since neither change you have here is required for provisioning virtual machines, it works as it is already.

I am touching parts that are needlessly complicating the configuration/setup for VMs. The whole part of having "base_" makes a sense for layers in containers but just makes the vm configuration more clunky, ugly and error-prone.
I wanted to work around this by creating VM-specific playbook that would replace playbook_image_service.yml and playbook_image_base.yml by You were strongly against this and wanted them to be included instead.

> 
>     * Red Hat must provide exactly the same content as other distributions. If you want to remove some things out to provide minimal testing environment, please do so with a switch that can be applied on all distributions so you have the same setup regardless of distribution.

Well I do strongly care about not having a ton of unneeded packages including build-chains and extra repositories (like gh-cli) on the machines as it taints the test environment. They should not be present there when the machine is not a client-devel (That is the switch You seek!). 
We do not use other distributions for verification so feel free to apply the changes to them on Your discretion.

> 
>     * The whole conversion from dnf to package is pointless. RHEL runs DNF, dnf package is installed on RHEL and yum is just symlink to dnf. Please avoid this change.
>
Well, not using distro-agnostic package module where possible just needlessly complicates this. 
It looks like I need to remind You that we are still officially supporting RHEL 6.10 and we need to run tests also on RHEL 7.X that does not have dnf but yum. 
 
>     * We don't have use for idempotency since you are not going to provision machine that was already setup, it just complicates stuff. For testing, you can just start from specific task with `--start-at-task `. If you insist on idempotency, then please, implement it on all tasks.

Idempotency is the main idea of ansible and we should be striving for it.
I expect that we might end-up rerunning some of the roles to fix a broken environment at some point.
I fixed what crashed on me when I was debugging stuff but I would like to revisit this in the future and improve the rest as well.

src/ansible/roles/packages/tasks/Fedora.yml
pbrezina: It is not present in master.

src/ansible/roles/packages/tasks/Fedora.yml
pbrezina: This is not present in master.

jakub-vavra-cz: This is present in master pulling build environment on IPA:
https://github.com/SSSD/sssd-ci-containers/blob/master/src/ansible/roles/packages/tasks/Fedora.yml#L133
I just added that it works on client and ipa without  "base_"

pbrezina:
Jakub, you are contributing to an open source project that has its vision and use cases. You can not just come and start changing things in an incompatible way to address your opinion and expect that it will be approved. This would not work for a foreign project and will not work in project owned by us either. When you contribute to any project, you have to follow the project coding style and patterns and make sure that you do not change existing functionality. As said, everything is the way it is for reasons. You can fix, you can improve but you can not break.

The whole purpose was to reuse playbooks used by this project to provision downstream virtual machines. This task was already accomplished. We can do stuff that you want, just do it without breaking things.

> > Hi Jakub, please, see some comments inside. And:
> > ```
> > * Do not introduce any functional or breaking changes, this repository is for containers and upstream use case first and things are the way they are for a reason. Especially since neither change you have here is required for provisioning virtual machines, it works as it is already.
> > ```
> 
> I am touching parts that are needlessly complicating the configuration/setup for VMs. The whole part of having "base_" makes a sense for layers in containers but just makes the vm configuration more clunky, ugly and error-prone. I wanted to work around this by creating VM-specific playbook that would replace playbook_image_service.yml and playbook_image_base.yml but You were strongly against this and You wanted them to be included instead.
> 
> > ```
> > * Red Hat must provide exactly the same content as other distributions. If you want to remove some things out to provide minimal testing environment, please do so with a switch that can be applied on all distributions so you have the same setup regardless of distribution.
> > ```
> 
> Well I do strongly care about not having a ton of unneeded packages including build-chains and extra repositories (like gh-cli) on the machines as it taints the test environment. They should not be present there when the machine is not a client-devel (That is the switch You seek!). We do not use other distributions for verification so feel free to apply the changes to them on Your discretion.

No, it is not the switch I seek, because we need it on other containers as well. The switch I seek would be "provision minimal testing environment" or "full environment" on every container/vm.

> > ```
> > * The whole conversion from dnf to package is pointless. RHEL runs DNF, dnf package is installed on RHEL and yum is just symlink to dnf. Please avoid this change.
> > ```
> 
> Well, not using distro-agnostic package module where possible just needlessly complicates this. It looks like I need to remind You that we are still officially supporting RHEL 6.10 and we need to run tests also on RHEL 7.X that does not have dnf but yum.

The Fedora package role will not work on RHEL6 and 7 so this comment is moot.

> > ```
> > * We don't have use for idempotency since you are not going to provision machine that was already setup, it just complicates stuff. For testing, you can just start from specific task with `--start-at-task `. If you insist on idempotency, then please, implement it on all tasks.
> > ```
> 
> Idempotency is the main idea of ansible and we should be striving for it. I expect that we might end-up rerunning some of the roles to fix a broken environment at some point. I fixed what crashed on me when I was debugging stuff but I would like to revisit this in the future and improve the rest as well.

Again, you must obey coding style of the project you are contributing to.

src/ansible/roles/packages/tasks/Fedora.yml
jakub-vavra-cz: Actually gdb, tcpdump and wireshark also do not make sense on all machines.
They got a specific usecases mostly on client.

pbrezina:
We had talk with Jakub, and we agreed on:

- idempotency yes, pbrezina will add tests for this to containers ci
- minimal test environment - yes, Jakub to create variable to switch between minimal and full installation, minimal will be default -> need a switch in containers ci
- ssh keys can be removed, document how to ignore host key check on localhost / add also pr to sssd-test-framework documentation (readme.md in sssd-ci-containers and https://tests.sssd.io/en/latest/running-tests.html in sssd-test-framework)
- do not put effort to Ubuntu and Debian as we will probably drop it in near future
- testing on RHEL7 and older to be discussed with QA and Alexey
- passkey testing - pbrezina to put passkey bits inside copr repository https://github.com/SSSD/sssd-ci-containers/issues/41
- packages required for sssd-test-framework,  required on the executor: https://github.com/SSSD/sssd-test-framework/blob/master/.github/workflows/ci.yml#L27 https://github.com/SSSD/sssd/blob/master/src/tests/system/requirements.txt
- packages required on client for the tests: ldb-tools, autofs, krb tools, tshark, tcpdump
- there is no other blocker to start running the system tests in downstream ci, polarion import will be implemented by idm-ops team

src/ansible/group_vars/all
```diff
@@ -56,3 +56,13 @@ user: {
...
+    server: [ freeipa-server, freeipa-server-dns, freeipa-server-trust-ad ],
+    client: [ freeipa-client, ]
+}
+
+ipa_packages: {
+    server: [ ipa-server-dns, ipa-server, ipa-server-trust-ad ],
+    client: [ ipa-client, ]
+}
```
ikerexxe, line 68: You are missing an empty line at the end of this file

src/ansible/roles/common/tasks/main.yml
```diff
@@ -65,9 +65,19 @@
     mode: '0600'
   with_items:
   - { src: 'root.id_rsa', dest: 'id_rsa' }
-  - { src: 'root.id_rsa.pub', dest: 'authorized_keys' }
+#  - { src: 'root.id_rsa.pub', dest: 'authorized_keys' }
```
ikerexxe, line 68: If you aren't going to use this line then it's better to remove it.

src/ansible/roles/packages/tasks/RedHat7.yml
```diff
@@ -0,0 +1,51 @@
+- name: Install packages for ground base image
+  block:
+  - name: Install systemd and minimal set of packages
+    package:
+      state: present
+      name:
+      - authconfig
+      - bind-utils
+      #- dbus-tools
```
ikerexxe, line 9: Same here, remove the line

src/ansible/roles/packages/tasks/RedHat7.yml
```diff
@@ -0,0 +1,51 @@
...
+      - expect
+      - firewalld
+      - openldap-clients
+      - openssh-clients
+      - openssh-server
+      - passwd
+      - policycoreutils
+      #- policycoreutils-python-utils
```
ikerexxe, line 17: Same here, remove the line

src/ansible/roles/packages/tasks/RedHat7.yml
```diff
@@ -0,0 +1,51 @@
...
+      - sssd
+      - sssd-*
+
+  - name: Install test dependencies on client
+    package:
+      state: present
+      name:
+      - ldb-tools
+      - net-tools
+      #- wireshark-cli
```
ikerexxe, line 46: Same here, remove the line

ikerexxe (commented):
Overall the changes seem fine but you forgot to add an empty line at the end of file in several places, and you also added new commented lines when it would be better to remove them.

Finally, the changes included in the `Code review fixes` commit would have been neater if you had made them in their respective commits, and not in a separate one.

src/ansible/group_vars/all
jakub-vavra-cz: Added.

src/ansible/roles/packages/tasks/RedHat7.yml
jakub-vavra-cz: Removed the whole redhat7 for now.

src/ansible/roles/common/tasks/main.yml
jakub-vavra-cz: Deleted.

src/ansible/roles/packages/tasks/RedHat7.yml
jakub-vavra-cz: Removed redhat7 for now.

src/ansible/roles/packages/tasks/Fedora.yml
jakub-vavra-cz: Moved to the extended package set.

src/ansible/roles/packages/tasks/Fedora.yml
jakub-vavra-cz: Moved to extended package set.

src/ansible/roles/dns/templates/etc.dnsmasq.conf.j2
```diff
@@ -0,0 +1,78 @@
+# This config is specific to downstream ci and does not
+# handle multiple AD DCs and other setups yet.
+# TODO: Add all AD servers
+# TODO: Remove meta_ip dependency (from  idm-ci).
+# TODO: Generalize this for all machines.
```
ikerexxe, line 5: If these TODOs were done then I think you should remove them

src/ansible/roles/packages/tasks/RedHat8.yml
```diff
@@ -0,0 +1,6 @@
+- name: Enable idm module
```
ikerexxe, line 1: I wonder, do we really want to decouple all Red Hat releases? If so, why didn't you define a file for RHEL7 and RHEL9?

src/ansible/roles/client/tasks/main.yml
```diff
@@ -82,3 +162,7 @@
     else
       exit 0
     fi
+
+- name: Show configured domains
+  debug:
+    msg: "domains = {{ ', '.join(domains) }}"
```
ikerexxe, line 168: Missing EOF

ikerexxe (commented):
Some minor comments

src/ansible/roles/client/tasks/main.yml
jakub-vavra-cz: Fixed.

src/ansible/roles/dns/templates/etc.dnsmasq.conf.j2
ikerexxe: Makes sense

src/ansible/roles/packages/tasks/RedHat8.yml
jakub-vavra-cz: This change was needed due to the magic done in the custom filter. 
The filter matches first distribution+number and then distribution without number so when there was RedHat and RedHat7 it always matched RedHat7.
When I put in RedHat7 it was incorrectly matching so Pavel said to do it this way.
I removed RHEL 7 from this MR to push forward but it might return in future.

ikerexxe: Understood. Thanks for the explanation

pbrezina:
I will review this on Monday, can you please rebase it in the mean time?

src/ansible/roles/ad/defaults/main.yml
```diff
@@ -0,0 +1,13 @@
+---
+# Users to be configured that their password never expires
+ad_permanent_users:
+- Administrator
+- vagrant
+# Skip vagrant-specific configuration of dns
+skip_dns: no
+# Skip installation of AD server
+skip_install: no
```
pbrezina, line 9: Maybe more specific name? skip_addc_install?

src/ansible/roles/ad/tasks/main.yml
```diff
@@ -84,32 +67,12 @@
   register: result
   failed_when: "result.rc != 255 and result.rc != 0"
   changed_when: "result.rc == 0"
+  until: "result.rc == 0"
```
pbrezina, line 70: Please, add a comment why retry is needed.

src/ansible/roles/client/tasks/main.yml
```diff
@@ -7,50 +11,126 @@
...
+      state: absent
+    with_items:
+    - /etc/krb5.conf
+    - /etc/krb5.keytab
+    - /etc/sssd/sssd.conf
+
+  - name: Add samba to domains
+    set_fact:
+      domains: "{{ domains + [service.samba.domain] }}"
+
```
pbrezina, line 89: Remove blank line.

src/ansible/roles/ldap/tasks/main.yml
```diff
@@ -27,6 +27,9 @@
       changetype: modify
       add: aci
       aci: (targetattr=*)(version 3.0; acl "Enable anyone read"; allow (read, search, compare)(userdn="ldap:///anyone");)
+  register: ldapmod
+  failed_when:
+    - 'ldapmod.rc != 0 and "ldap_modify: Type or value exists" not in ldapmod.stderr'
```
pbrezina, line 32: Deindent.

src/ansible/roles/packages/tasks/RedHat8.yml
```diff
@@ -0,0 +1,6 @@
...
+  command: yum module enable idm:DL1 -y
+  when: ansible_distribution_major_version == '8'
```
pbrezina, line 3: when is not needed here.

src/ansible/roles/packages/tasks/RedHat8.yml
pbrezina, line 1: I still think you want to include CentOS8.yml here. You limit the packages that are installed with "extended_packageset". If that is set to True then EPEL and sssd-ci-deps repo will be missing.

src/ansible/roles/ssh_server/tasks/main.yml
```diff
@@ -11,3 +11,7 @@
...
+  until: "restart_res is not failed"
```
pbrezina, line 15: Please, add comment why restart is needed.

pbrezina (changes requested):
Hi, looks good to me, there are few minor comments.

Now, it would be good, if you can approve https://github.com/SSSD/sssd-ci-containers/pull/51 before we merge this one and incorporate it into this pull request so passkey testing functionality is always built. It moves the bits into stand alone rpm package so build dependencies are no longer required to be installed.

I'm running the provisioning also locally so I can check that all existing tests pass. I'll get back with results.

src/ansible/roles/ssh_server/tasks/main.yml
jakub-vavra-cz: Done.

src/ansible/roles/ldap/tasks/main.yml
jakub-vavra-cz: done

src/ansible/roles/client/tasks/main.yml
jakub-vavra-cz: done

src/ansible/roles/ad/tasks/main.yml
jakub-vavra-cz: done

src/ansible/roles/ad/defaults/main.yml
jakub-vavra-cz: done

src/ansible/roles/packages/tasks/RedHat8.yml
jakub-vavra-cz: Added epel for extended packageset. We should be god now. 
The ci-deps will be in place with https://github.com/SSSD/sssd-ci-containers/pull/51

src/ansible/roles/ad/tasks/main.yml
```diff
@@ -14,60 +15,42 @@
...
-  register: schema
-  failed_when: schema.rc != 0 and schema.stdout is not search('ENTRY_EXISTS')
-  changed_when: schema.rc == 0
-  with_items:
-  - sudo
+- name: 'Add sudo schema and possibly other'
+  include_tasks: 'schema.yml'
+  when: not skip_schema
 
 - name: Set Password Never Expires for system users
```
pbrezina, line 55: This step has never ones failed before and now it takes 5 retries for each account. Please, debug it. It takes really long and it can not be because AD is not responsive, because each account takes 5 retries so it has to retry even when we were already able to modify the first user. Maybe try to reboot just before "Make sure Active Directory Web Services is running" step?

pbrezina (changes requested):
LGTM, tests pass, AD can be now provisioned again: But it would really be good to solve the "password never expires" step. There's something wrong going on, the retry should not be needed.

src/ansible/roles/ad/tasks/main.yml
jakub-vavra-cz: I have created an issue for this https://github.com/SSSD/sssd-ci-containers/issues/52.
I will look in into it as a part of AD improvements. Please do not block this MR as others are waiting for it.

src/ansible/roles/packages/tasks/RedHat8.yml
jakub-vavra-cz: It is needed after all on rhel 9:
fatal: [master.ipa.test]: FAILED! => changed=true 
  cmd:
  - yum
  - module
  - enable
  - idm:DL1
  - -y
  delta: '0:00:02.163913'
  end: '2023-07-25 06:13:18.094871'
  msg: non-zero return code
  rc: 1
  start: '2023-07-25 06:13:15.930958'
  stderr: |-
    Error: Problems in request:
    missing groups or modules: idm:DL1
  stderr_lines: <omitted>
  stdout: |-
    rhel-9.3 buildroot                               57 MB/s | 3.6 MB     00:00
    rhel-9.3 buildroot debuginfo                     56 MB/s | 2.5 MB     00:00
  stdout_lines: <omitted>

pbrezina (approved):

PR #22, opened 2023-02-09
matrix: prefer rawhide over numbered version

pbrezina:
The latest version may not yet be available as container when
radhide branch.


pbrezina:
The problem was that latest devel branch of Fedora was not able to run in containers so either way, we would have to wait for it to be fixed. This patch should not be needed.

PR #24, opened 2023-02-20
Make sssd-ci-containers compatible with idm-ci provisioning

pbrezina:


pbrezina:
CI fails because the renamed certificates are not yet available in quay. This will be fixed automatically when merged.

PR #25, opened 2023-02-20
compose: add SYS_CHROOT capability

ikerexxe:
ssh connections are failing due to:
fatal: chroot("/usr/share/empty.sshd"): Operation not permitted [preauth]

Adding SYS_CHROOT capability fixes the problem. This solution is based on https://access.redhat.com/solutions/3626531

Unfortunately I don't have time to investigate the issue more profoundly. I don't know why the problem appeared now and it was working correctly before. If you don't like the solution, take this patch as a problem report and provide a better fix.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -9,6 +9,7 @@ services:
     cap_add:
     - NET_RAW
     - NET_ADMIN
+    - SYS_CHROOT
     security_opt:
     - apparmor=unconfined
     - label=disable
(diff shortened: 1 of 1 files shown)


pbrezina:
Does it happen with some specific version?

ikerexxe:
I noticed it today, but this might have been happening for a while. I hadn't updated the images for a while (circa November) and the branch I was using was latest updated from master the January 25, and the problem happened. So I updated both things and the problem persisted, that's when I started looking for a solution.

pbrezina:
The issue started to happen when I upgraded to F37. Ack.

PR #26, opened 2023-02-28
Add passkey testing support

ikerexxe:
The following packages, capabilities and binaries are needed to perform the passkey testing in an automated way.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,6 +9,11 @@ push:
 up:
 	docker-compose up --detach ${LIMIT}
 
+up-passkey:
+	export HIDRAW=$(shell fido2-token -L|cut -f1 -d:) \
+	&& docker-compose -f docker-compose.yml -f docker-compose.passkey.yml up \
+	--detach ${LIMIT}
+
 stop:
 	docker-compose stop
 
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -30,6 +30,7 @@ services:
     - SYS_PTRACE
     - AUDIT_WRITE
     - SYS_CHROOT
+    - NET_ADMIN
     security_opt:
     - apparmor=unconfined
     - label=disable
--- a/src/ansible/roles/client/tasks/main.yml
+++ b/src/ansible/roles/client/tasks/main.yml
@@ -82,3 +82,24 @@
     else
       exit 0
     fi
+
+- name: Passkey testing
+  block:
+  - name: Copy random.c to /opt
+    copy:
+      src: '{{ playbook_dir }}/roles/client/files/random.c'
+      dest: /opt/random.c
+      owner: root
+      group: root
+      mode: 0644
+
+  - name: Build random.so
+    shell: |
+      cd /opt
+      gcc -fPIC -shared -o random.so random.c -lcrypto
+
+  - name: Remove random.c
+    file:
+      path: /opt/random.c
+      state: absent
+  when: passkey_support
--- a/src/ansible/roles/facts/tasks/CentOS.yml
+++ b/src/ansible/roles/facts/tasks/CentOS.yml
@@ -1,2 +1,0 @@
-- name: 'Facts are the same as in Fedora'
-  include_tasks: 'Fedora.yml'
--- a/src/ansible/roles/packages/tasks/Fedora.yml
+++ b/src/ansible/roles/packages/tasks/Fedora.yml
@@ -75,6 +75,14 @@
         libtalloc               \
         libtevent
     when: debuginfo
+  - name: Install packages required for passkey testing
+    dnf:
+      state: present
+      name:
+      - gcc
+      - openssl-devel
+      - umockdev
+    when: passkey_support
   when: "'base_client' in group_names"
 
 - name: Install packages for LDAP base image
--- a/src/ansible/roles/packages/tasks/Ubuntu.yml
+++ b/src/ansible/roles/packages/tasks/Ubuntu.yml
@@ -49,6 +49,14 @@
       - realmd
       - sssd
       - sssd-*
+  - name: Install packages required for passkey testing
+    apt:
+      state: present
+      name:
+      - gcc
+      - libssl-dev
+      - umockdev
+    when: passkey_support
   when: "'base_client' in group_names"
 
 - name: Install packages for NFS base image
@@ -132,7 +140,6 @@
       - libselinux1-dev
       - libsemanage-dev
       - libsmbclient-dev
-      - libssl-dev
       - libssl-dev
       - libsystemd-dev
       - libtalloc-dev
--- a/docker-compose.passkey.yml
+++ b/docker-compose.passkey.yml
@@ -1,0 +1,6 @@
+services:
+  client:
+    cap_add:
+    - SYS_RAWIO
+    devices:
+    - "$HIDRAW:$HIDRAW"
--- a/src/ansible/roles/client/files/random.c
+++ b/src/ansible/roles/client/files/random.c
@@ -1,0 +1,151 @@
+#undef NDEBUG
+#include <assert.h>
+#include <stdlib.h>
+#include <stdio.h>
+#include <string.h>
+
+#include <openssl/crypto.h>
+#include <openssl/rand.h>
+#include <openssl/provider.h>
+#include <openssl/core_names.h>
+
+static OSSL_FUNC_rand_newctx_fn mock_rand_newctx;
+static OSSL_FUNC_rand_freectx_fn mock_rand_freectx;
+static OSSL_FUNC_rand_instantiate_fn mock_rand_instantiate;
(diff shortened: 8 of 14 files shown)


ikerexxe:
@madhuriupadhye, @spoore1 these are the changes needed to run the passkey testing in an automated way. It allows to playback the recorded umockdev data, but not to record it. For that purpose additional changes are needed and I don't think they should be included in this repository/branch. If you still need to record anything please use https://github.com/ikerexxe/sssd-ci-containers/tree/passkey.

docker-compose.yml
```diff
@@ -91,6 +92,7 @@ services:
     - NET_ADMIN
     - AUDIT_WRITE
     - SYS_CHROOT
+    #- SYS_RAWIO # Only needed if you are using real FIDO2 tokens
```
pbrezina, line 95: This should be in different commit and perhaps use "only needed if you require access to real devices like FIDO2 tokens"?

src/ansible/roles/packages/tasks/Fedora.yml
```diff
@@ -74,6 +74,15 @@
...
         libtevent
+  - name: Install packages required for passkey testing
+    dnf:
+      state: present
+      name:
+        - gcc
+        - openssl-devel
+        - umockdev
+    when: (ansible_facts['distribution'] == "CentOS" and ansible_distribution_major_version | int >= 9) or
+          (ansible_facts['distribution'] == "Fedora")
```
pbrezina, line 85: It would be better to add new fact to https://github.com/SSSD/sssd-ci-containers/tree/master/src/ansible/roles/facts/tasks

E.g.

```yaml
passkey_support: Yes|No
```

End then use just `when: passkey_support`

src/ansible/roles/client/tasks/main.yml
```diff
@@ -82,3 +82,22 @@
...
     fi
+
+- name: Copy random.c to /opt
+  copy:
+    src: '{{ playbook_dir }}/../tools/random.c'
+    dest: /opt/random.c
+    owner: root
+    group: root
+    mode: 0644
+  when: (ansible_facts['distribution'] == "Fedora") or
```
pbrezina, line 93: Use new fact as well.

src/ansible/roles/client/tasks/main.yml
```diff
@@ -82,3 +82,22 @@
...
+        (ansible_facts['distribution'] == "Ubuntu") or
+        (ansible_facts['distribution'] == "CentOS" and ansible_distribution_major_version | int >= 9)
+        
+- name: Build random.so needed for passkey testing
+  shell: |
+    cd /opt
+    gcc -fPIC -shared -o random.so random.c -lcrypto
+  when: (ansible_facts['distribution'] == "Fedora") or
```
pbrezina, line 101: New fact as well.

src/ansible/roles/client/tasks/main.yml
pbrezina: The file should be removed after compilation.

You can also wrap it with `block` together with compilation.

pbrezina (changes requested):
Thank you, see minor comments inside.

pbrezina:
> @madhuriupadhye, @spoore1 these are the changes needed to run the passkey testing in an automated way. It allows to playback the recorded umockdev data, but not to record it. For that purpose additional changes are needed and I don't think they should be included in this repository/branch. If you still need to record anything please use https://github.com/ikerexxe/sssd-ci-containers/tree/passkey.

What additional changes are needed? Maybe we can include the as well so we can easily create new recordings in the future on newer containers.

ikerexxe:
We would need to add the `SYS_RAWIO` capability, and find a way to export the exact device where the fido2 token is connected (the exact line changes are linked below). I don't know if those changes are compatible with a cloud environment as they may be used to access low-level stuff.

Changes:
https://github.com/ikerexxe/sssd-ci-containers/blob/passkey/Makefile#L1
https://github.com/ikerexxe/sssd-ci-containers/blob/passkey/Makefile#L12
https://github.com/ikerexxe/sssd-ci-containers/blob/passkey/docker-compose.yml#L35

src/ansible/roles/packages/tasks/Fedora.yml
ikerexxe: I hope I did it correctly. Will need to check the results later.

By the way, thanks for the lesson about facts ;)

src/ansible/roles/client/tasks/main.yml
ikerexxe: Done both of the improvements

src/ansible/roles/packages/tasks/Fedora.yml
```diff
@@ -74,6 +74,14 @@
...
         libtalloc               \
         libtevent
+  - name: Install packages required for passkey testing
+    dnf:
+      state: present
+      name:
+        - gcc
+        - openssl-devel
+        - umockdev
+    when: passkey_support
```
pbrezina, line 84: Please align the list to the key, ie.

```
key:
- value
- value
```

src/ansible/roles/client/tasks/main.yml
```diff
@@ -82,3 +82,24 @@
     else
       exit 0
     fi
+
+- name: Passkey testing
+  block:
+  - name: Copy random.c to /opt
+    copy:
+      src: '{{ playbook_dir }}/../tools/random.c'
```
pbrezina, line 90: You can put this to files folder in this role.

pbrezina (changes requested):
Thank you, two more nitpicks and we can merge.

src/ansible/roles/packages/tasks/Ubuntu.yml
```diff
@@ -49,6 +49,13 @@
       - realmd
       - sssd
       - sssd-*
+  - name: Install packages required for passkey testing
+    apt:
+      state: present
+      name:
+      - gcc
+      - libssl-dev
+      - umockdev
```
pbrezina, line 58: when: passkey_support to be consistent with Fedora?

docker-compose.yml
```diff
@@ -91,6 +92,7 @@ services:
     - NET_ADMIN
     - AUDIT_WRITE
     - SYS_CHROOT
+    #- SYS_RAWIO # Only needed if you require access to real devices like FIDO2 tokens
```
pbrezina, line 95: If this can not work on its own, I think we can remove this patch.

pbrezina (changes requested):

ikerexxe:
I added the override and the new target in the Makefile

src/ansible/roles/packages/tasks/Ubuntu.yml
ikerexxe: Done

docker-compose.yml
ikerexxe: I removed this patch

src/ansible/roles/packages/tasks/Ubuntu.yml
```diff
@@ -132,8 +140,6 @@
       - libselinux1-dev
       - libsemanage-dev
       - libsmbclient-dev
-      - libssl-dev
-      - libssl-dev
```
pbrezina, line 136: OpenSSL devel is required to build SSSD. This should stay, since it is installed only conditionally for passkey.

Makefile
```diff
@@ -1,3 +1,5 @@
+SHELL_OUT = $(shell fido2-token -L|cut -f1 -d:)
```
pbrezina, line 1: Can this be moved to up-passkey?

src/ansible/roles/packages/tasks/Ubuntu.yml
ikerexxe: Added it back

Makefile
ikerexxe: Done

src/ansible/roles/facts/tasks/CentOS.yml
```diff
@@ -1,2 +1,19 @@
...
+
+- name: Define passkey for CentOS9+
+  set_fact:
+    passkey_support: Yes
+  when: (ansible_distribution_major_version | int >= 9)
+
+- name: Define passkey for CentOS8-
+  set_fact:
+    passkey_support: No
+  when: (ansible_distribution_major_version | int < 9)
```
pbrezina, line 19: You can keep the Fedora include and then just override what you need (see https://github.com/SSSD/sssd-ci-containers/blob/master/src/ansible/roles/facts/tasks/RedHat.yml). It may also be clearer to separate this to CentOS8.yml and CentOS9.yml.

pbrezina:
Thank you. Ack.

pbrezina (approved):

PR #31, opened 2023-03-22
Makefile: Fix option ordering for passkey-up target

justin-stephenson:


sumit-bose:
Hi,

thanks, works for me. I  wonder, since `--detach` is already move to the next line, if you would like to move `--no-recreate` to the next line as well to keep the line shorter in the editor? Otherwise ACK.

bye,
Sumit

sumit-bose (approved):

justin-stephenson:
> 

Makes sense, fixed.

sumit-bose:
Thanks, still ACK.

PR #35, opened 2023-03-29
fedora: install e2fsprogs

ikerexxe:
https://github.com/SSSD/sssd/commit/c6db359fa8b75ed0d03b7211c588e8c317d80c26 added a new dependency for the multihost tests to the chattr command, which is provided by the e2fsprogs package.

### DIFF
--- a/src/ansible/roles/packages/tasks/Fedora.yml
+++ b/src/ansible/roles/packages/tasks/Fedora.yml
@@ -170,6 +170,12 @@
       - uid_wrapper
       - valgrind
 
+  - name: Install packages required for multihost tests
+    dnf:
+      state: present
+      name:
+      - e2fsprogs
+
   - name: Install additional python packages
     pip:
       name:
--- a/src/ansible/roles/packages/tasks/Ubuntu.yml
+++ b/src/ansible/roles/packages/tasks/Ubuntu.yml
@@ -170,6 +170,13 @@
       - xml-core
       - xsltproc
 
+  - name: Install packages required for multihost tests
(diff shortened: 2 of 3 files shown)


pbrezina:
Can you please also add it to Ubuntu and Debian to keep the content the same?

ikerexxe:
Done, but strictly speaking we don't need to explicitly install `e2fsprogs` for those distributions because it comes in the base images.

src/ansible/roles/packages/tasks/Ubuntu.yml
```diff
@@ -170,6 +170,12 @@
       - xml-core
       - xsltproc
 
+  - name: Install packages required for multihost tests
+    dnf:
```
pbrezina, line 174: apt

src/ansible/roles/packages/tasks/Fedora.yml
```diff
@@ -170,6 +170,12 @@
       - uid_wrapper
       - valgrind
 
+  - name: Install packages required for multihost tests
+    dnf:
```
pbrezina, line 174: apt

pbrezina (changes requested):
Also please change commit subject to "packages: add e2fsprogs"

src/ansible/roles/packages/tasks/Fedora.yml
ikerexxe: Fixed

pbrezina (approved):

PR #36, opened 2023-03-30
Add Keycloak Integration test containers

spoore1:
Adding new container for Keycloak.  Container defined in separate docker-compose since it is unnecessary for general SSSD testing.


Makefile
```diff
@@ -14,12 +14,19 @@ up-passkey:
...
+	--no-recreate --detach ${LIMIT}
+
 stop:
 	docker-compose stop
 
 down:
 	docker-compose down
 
+down-keycloak:
+	docker-compose -f docker-compose.yml -f docker-compose.keycloak.yml down
```
ikerexxe, line 28: Do we really need another target to bring it down? I guess it's because you are adding a new container but I want to make sure

src/ansible/roles/keycloak/defaults/main.yml
```diff
@@ -0,0 +1,7 @@
+keycloak_name: "keycloak"
+#keycloak_version: "19.0.3"
+keycloak_version: "17.0.1"
```
ikerexxe, line 3: Why version `17` when upstream released version `21` two months ago? I'm not following keycloak releases profoundly but I'd expect to use the latest version (`21`) or the previous one (`20`)

src/ansible/roles/keycloak/tasks/main.yml
```diff
@@ -0,0 +1,134 @@
...
+
+      [Install]
+      WantedBy=multi-user.target
+    dest: /etc/systemd/system/keycloak.service
+
+- name: Reload systemd to pickup changes added for keycloak
+  systemd:
+    daemon_reload: yes
+
+- name: Run build step for Keycloak and RH-SSO 8+ (skip for RH-SSO 7)
```
ikerexxe, line 116: The description mentions that RH-SSO 7 should be skipped but I don't see any filtering for this command. I guess you can delete that part of the description as we aren't producing any image for RH-SSO 7

src/ansible/roles/keycloak/tasks/main.yml
```diff
@@ -0,0 +1,134 @@
...
+  shell: |
+    su - keycloak -c '''
+    export KEYCLOAK_ADMIN=admin
+    export KEYCLOAK_ADMIN_PASSWORD=Secret123
```
ikerexxe, line 120: If possible I'd try to avoid defining the secrets and passwords in the tasks and do that in the [proper file](https://github.com/SSSD/sssd-ci-containers/blob/master/src/ansible/group_vars/all#L7)

src/ansible/roles/keycloak/tasks/main.yml
```diff
@@ -0,0 +1,134 @@
...
+    export KC_HOSTNAME=$(hostname):8443
+    export KC_HTTPS_CERTIFICATE_FILE=/data/certs/master.keycloak.test.crt
+    export KC_HTTPS_CERTIFICATE_KEY_FILE=/data/certs/master.keycloak.test.key
+    export KC_HTTPS_TRUST_STORE_FILE=/data/certs/master.keycloak.test.keystore
+    export KC_HTTPS_TRUST_STORE_PASSWORD=Secret123
```
ikerexxe, line 125: Another password here

ikerexxe (changes requested):
Some minor issues and questions inline

src/ansible/roles/keycloak/defaults/main.yml
spoore1: 17 and 19 are just the versions that we've been testing the new SCIM user storage plugin with.   I'll look into locking it down to a newer version but, we will need to manage that versioning somehow so we can know which version of KC is in the container.  Thoughts?

ikerexxe: I think it's fine to keep the reference to the keycloak version in use, but as mentioned previously I'd expect it to be one of the newest ones.

Makefile
spoore1: What if I modify the down target to down using all the docker-compose files?

src/ansible/roles/keycloak/defaults/main.yml
spoore1: OK, I set the var to the current latest.  removing commented ones in the upcoming push

Makefile
ikerexxe: That's fine for me, but does it work correctly if we aren't using any override?

spoore1: Seems to:

I first ran make up (no overrides just default docker-compose.yml.   This is all that was running:

```
# podman ps
CONTAINER ID  IMAGE                          COMMAND     CREATED        STATUS        PORTS       NAMES
2c9465cd0f1e  quay.io/sssd/ci-samba:latest   /sbin/init  3 minutes ago  Up 3 minutes              samba
effe4d655585  quay.io/sssd/ci-client:latest  /sbin/init  3 minutes ago  Up 3 minutes              client
07ba4196ab25  quay.io/sssd/ci-dns:latest                 3 minutes ago  Up 3 minutes              dns
771946867f3c  quay.io/sssd/ci-ldap:latest    /sbin/init  3 minutes ago  Up 3 minutes              ldap
e5cc9596b515  quay.io/sssd/ci-ipa:latest     /sbin/init  3 minutes ago  Up 3 minutes              ipa
57ebc7dfde51  quay.io/sssd/ci-kdc:latest     /sbin/init  3 minutes ago  Up 3 minutes              kdc
2f82dfb1cd88  quay.io/sssd/ci-nfs:latest     /sbin/init  3 minutes ago  Up 3 minutes              nfs
```

Then I ran make down using all 3 files:

```
# make down
docker-compose -f docker-compose.yml \
-f docker-compose.keycloak.yml \
-f docker-compose.passkey.yml down
WARNING: The HIDRAW variable is not set. Defaulting to a blank string.
Stopping dns    ... done
Stopping samba  ... done
Stopping nfs    ... done
Stopping kdc    ... done
Stopping ldap   ... done
Stopping ipa    ... done
Stopping client ... done
Removing dns    ... done
Removing samba  ... done
Removing nfs    ... done
Removing kdc    ... done
Removing ldap   ... done
Removing ipa    ... done
Removing client ... done
Removing network sssd-ci
```

On RHEL8 with podman-compose from pip, it does throw some errors but exits with rc 0.   And it does that for any/all containers that are not up not just the additional ones from the override files.

ikerexxe (approved):
LGTM. Thanks for your great work!

src/ansible/roles/keycloak/tasks/main.yml
```diff
@@ -0,0 +1,148 @@
+- name: print keycloak_url
```
pbrezina, line 1: Uppercase P

src/ansible/roles/keycloak/tasks/main.yml
```diff
@@ -0,0 +1,148 @@
...
+    dest: /tmp
+
+- name: Check if unzipped directory exists
+  stat:
+    path: "{{ unzip_dir }}"
+  register: unzip_dir_chk
+
+- name: debug unzip dir
+  debug:
+    var: unzip_dir_chk.stat
```
pbrezina, line 23: This is probably not needed. If unzip succeeded, you know it exist.

src/ansible/roles/keycloak/tasks/main.yml
```diff
@@ -0,0 +1,148 @@
...
+- name: Download software zipfile
+  get_url:
+    url: "{{ keycloak_url }}"
+    dest: /tmp
+    
+- name: Unzip software
+  unarchive:
+    remote_src: yes
+    src: "/tmp/{{ zip_file }}"
+    dest: /tmp
```
pbrezina, line 14: I believe there is a way to omit the root folder from the zip file and set custom name, so you can extract it to /tmp/keycloak instead of a versioned folder.

src/ansible/roles/keycloak/tasks/main.yml
```diff
@@ -0,0 +1,148 @@
...
+    var: unzip_dir_chk.stat
+
+- name: Check if /opt/keycloak dir exists
+  stat:
+    path: /opt/keycloak
+  register: opt_keycloak_chk
+
+- name: debug opt dir
+  debug:
+    var: opt_keycloak_chk.stat
```
pbrezina, line 32: I think this is not needed.

src/ansible/roles/keycloak/tasks/main.yml
```diff
@@ -0,0 +1,148 @@
...
+
+- name: Move software to /opt
+  command: mv {{ unzip_dir }} /opt/keycloak 
+  when: unzip_dir_chk.stat.exists and not opt_keycloak_chk.stat.exists
```
pbrezina, line 36: You can extract it directly to /opt/keycloak.

Also IMHO you don't need to care about impotence here. You can just override the folder if it exist - which it never does in containers.

src/ansible/roles/keycloak/tasks/main.yml
```diff
@@ -0,0 +1,148 @@
...
+
+- name: create Keycloak user/group
```
pbrezina, line 38: Uppercase C

src/ansible/roles/keycloak/tasks/main.yml
```diff
@@ -0,0 +1,148 @@
...
+      -trustcacerts -storepass {{ service.keycloak.admin_password }}
+
+- name: Copy CA Certificate to local pki anchors
+  copy:
+    src: /data/certs/ca.crt
+    dest: "{{ ca_trust_dir }}"
+    remote_src: yes
+
+- name: Update system ca trusts to pick up new certificate
+  command: "{{ ca_trust_update }}"
```
pbrezina, line 84: This can go to the common role.

src/ansible/roles/keycloak/tasks/main.yml
```diff
@@ -0,0 +1,148 @@
...
+
+- name: create /etc/sysconfig if it doesn't exist
+  file:
+    name: /etc/sysconfig
+    state: directory
```
pbrezina, line 89: Is this needed?

src/ansible/roles/keycloak/tasks/main.yml
```diff
@@ -0,0 +1,148 @@
...
+    export KEYCLOAK_ADMIN_PASSWORD={{ service.keycloak.admin_password }}
+    export KC_HOSTNAME=$(hostname):8443
+    export KC_HTTPS_CERTIFICATE_FILE=/data/certs/master.keycloak.test.crt
+    export KC_HTTPS_CERTIFICATE_KEY_FILE=/data/certs/master.keycloak.test.key
+    export KC_HTTPS_TRUST_STORE_FILE=/data/certs/master.keycloak.test.keystore
+    export KC_HTTPS_TRUST_STORE_PASSWORD={{ service.keycloak.admin_password }}
+    export KC_HTTP_RELATIVE_PATH=/auth
+    /opt/keycloak/bin/kc.sh build
+    '''
+  register: build_out
```
pbrezina, line 143: build_out is unused

src/build.sh
```diff
@@ -121,8 +121,10 @@ if [ "$SKIP_BASE" == 'no' ]; then
   build_base_image "ci-base-ldap:${TAG}"   base-ipa
   build_base_image "ci-base-ground:${TAG}" base-nfs
   build_base_image "ci-base-ground:${TAG}" base-kdc
+  build_base_image "ci-base-ground:${TAG}" base-keycloak
 fi
 
+set -e
```
pbrezina, line 127: This shouldn't be needed, there's set -xe on the top of the file.

pbrezina (changes requested):
Thank you, please see comments inside. The comments are about:

* Keep it simple. Keep in mind that the playbook has specific use case so it does not have to be perfectly sharable with others. We can afford to make some assumptions. The more simple the playbook is, the more simple it is to maintain it.
* Start task names with uppercase letters 

It would be also good to detect latest keycloak release instead of hardcoding a specific version. See [this](https://github.com/SSSD/sssd-ci-containers/blob/master/src/Containerfile#L24) for inspiration.

src/build.sh
spoore1: removed

src/ansible/roles/keycloak/tasks/main.yml
spoore1: removed

src/ansible/roles/keycloak/tasks/main.yml
spoore1: I thought I had added it for a specific reason but, it does not appear to be needed.   Removed.

src/ansible/roles/keycloak/tasks/main.yml
spoore1: moved to common

src/ansible/roles/keycloak/tasks/main.yml
```diff
@@ -0,0 +1,148 @@
...
+    owner: keycloak
+    group: keycloak
+    recurse: yes
+
+- name: change ownership of files in /data/certs
+  file:
+    path: /data/certs
+    owner: keycloak
+    group: keycloak
+    recurse: yes
```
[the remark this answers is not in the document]
spoore1: Keycloak needs to be able to read the key as the keycloak user with limited permissions.   I modified the change to only change the cert and key owner and group.

src/ansible/roles/keycloak/tasks/main.yml
spoore1: I think I changed all the task names to start with uppercase.

src/ansible/inventory.yml
```diff
@@ -62,6 +65,11 @@ all:
           hosts:
             kdc.test:
               ansible_host: sssd-wip-kdc
+
```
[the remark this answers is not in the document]
spoore1: removed

data/configs/dnsmasq.conf
```diff
@@ -20,6 +20,7 @@ address=/master.ldap.test/172.16.100.20
 address=/client.test/172.16.100.40
 address=/nfs.test/172.16.100.50
 address=/kdc.test/172.16.100.60
+address=/master.keycloak.test/172.16.100.11
```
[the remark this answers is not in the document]
spoore1: changed the IP to .70.

src/ansible/roles/keycloak/tasks/main.yml
spoore1: Extracts directly to dir now.

src/ansible/roles/keycloak/tasks/main.yml
spoore1: fixed

src/ansible/roles/keycloak/tasks/main.yml
```diff
@@ -0,0 +1,148 @@
...
+  when: unzip_dir_chk.stat.exists and not opt_keycloak_chk.stat.exists
+
+- name: create Keycloak user/group
+  user:
+    name: keycloak
+    home: /opt/keycloak
+    system: yes
+    create_home: no
+
+- name: change ownership of files in /opt/keycloak
```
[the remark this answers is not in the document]
spoore1: done

ikerexxe:
There's a problem in CentOS 8 and Debian derivatives with the `Get keycloak tags` task.

docker-compose.keycloak.yml
```diff
@@ -0,0 +1,18 @@
...
+  keycloak:
+    image: ${REGISTRY}/ci-keycloak:${TAG}
+    container_name: keycloak
+    hostname: master.keycloak.test
+    dns: 172.16.100.2
+    env_file: ./env.containers
+    cap_add:
+    - SYS_ADMIN
+    - SYS_PTRACE
+    - AUDIT_WRITE
```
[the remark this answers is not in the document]
pbrezina: Ok. Can you please add AUDIT_CONTROL to docker-compose.yml in standalone commit with a description of why it is needed? Otherwise we will forget it.

src/ansible/roles/common/tasks/main.yml
```diff
@@ -90,3 +90,12 @@
   - { src: 'ci.id_rsa', dest: 'id_rsa' }
   - { src: 'ci.id_rsa.pub', dest: 'authorized_keys' }
   - { src: 'ci.id_rsa.pub', dest: 'id_rsa.pub' }
+
+- name: Copy CA Certificate to local pki anchors
```
pbrezina, line 94: certificate with lowercase c

src/ansible/roles/keycloak/tasks/main.yml
```diff
@@ -0,0 +1,110 @@
...
+
+- name: Unzip software
+  unarchive:
+    remote_src: yes
+    src: "/tmp/{{ zip_file }}"
+    dest: /opt/keycloak
+    owner: keycloak
+    group: keycloak
+    extra_opts:
+      - --strip-components=1
```
pbrezina, line 25: Align - with extra_opts:

```yml
extra_opts:
- --strip-components=1
```

pbrezina: This applies everywhere.

src/ansible/roles/keycloak/tasks/main.yml
```diff
@@ -0,0 +1,110 @@
...
+
+- name: Change ownership of files in /data/certs
+  file:
+    path: "/data/certs/{{ item }}"
+    owner: keycloak
+    group: keycloak
+  loop:
+    - master.keycloak.test.crt
+    - master.keycloak.test.key
```
pbrezina, line 34: crt is already readable by everyone, I think we can make key readable as well in common.

pbrezina: that would render this step unnecessary, unless you need the ownership.

src/ansible/roles/keycloak/tasks/main.yml
```diff
@@ -0,0 +1,110 @@
...
+    su - keycloak -c '''
+    export KEYCLOAK_ADMIN=admin
+    export KEYCLOAK_ADMIN_PASSWORD={{ service.keycloak.admin_password }}
+    export KC_HOSTNAME=$(hostname):8443
+    export KC_HTTPS_CERTIFICATE_FILE=/data/certs/master.keycloak.test.crt
+    export KC_HTTPS_CERTIFICATE_KEY_FILE=/data/certs/master.keycloak.test.key
+    export KC_HTTPS_TRUST_STORE_FILE=/data/certs/master.keycloak.test.keystore
+    export KC_HTTPS_TRUST_STORE_PASSWORD={{ service.keycloak.admin_password }}
+    export KC_HTTP_RELATIVE_PATH=/auth
+    /opt/keycloak/bin/kc.sh build
```
pbrezina, line 104: Doesn't it make sense to build keycloak before you create the systemd unit?

src/ansible/roles/keycloak/tasks/main.yml
```diff
@@ -0,0 +1,110 @@
...
+    create_home: yes
+
+- name: Print keycloak_url
+  debug:
+    msg: "keycloak_url={{ keycloak_url }}"
+
+- name: Download software zipfile
+  get_url:
+    url: "{{ keycloak_url }}"
+    dest: /tmp
```
pbrezina, line 15: You can use /tmp/keycloak.tgz or other file name, this will make the variables unneeded. Then the only fact you need is the keycloak_url, please remove it from facts and put it to this role.

pbrezina (changes requested):

docker-compose.keycloak.yml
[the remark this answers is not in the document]
spoore1: Should I pull it out and add it as a separate PR too or just another commit in this PR?

src/ansible/roles/keycloak/tasks/main.yml
spoore1: I'll double check the ownership and permissions requirement.

spoore1: @pbrezina 

I tried changing all key file permissions to 644 but, that breaks the ipa trust-add to the samba domain.   It makes more sense to me to adjust the ownership or permissions for the file in the keycloak role since that's the only one that requires a different setting.  

Do you still prefer I change it in common or can it stay in the keycloak role?   Or do you have another suggestion?

src/ansible/roles/keycloak/tasks/main.yml
spoore1: Changed.

src/ansible/roles/keycloak/tasks/main.yml
spoore1: Moved.

src/ansible/roles/keycloak/tasks/main.yml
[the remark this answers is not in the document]
spoore1: marking this one resolved and we can continue this in the other comment thread.

src/ansible/roles/keycloak/tasks/main.yml
spoore1: Changed

src/ansible/roles/common/tasks/main.yml
spoore1: fixed.

src/ansible/roles/keycloak/tasks/main.yml
pbrezina: Ok, if it breaks, then keep it.

src/ansible/roles/keycloak/tasks/main.yml
```diff
@@ -0,0 +1,119 @@
...
+    keycloak_url: "{{ base_url }}/{{ result.json[1].name }}/keycloak-{{ result.json[1].name }}.tar.gz"
+
+- name: Create Keycloak user/group
+  user:
+    name: keycloak
+    home: /opt/keycloak
+    system: yes
+    create_home: yes
+
+- name: Print keycloak_url
```
pbrezina, line 21: Please, move this after set_fact so it is togther.

pbrezina (changes requested):
One last minor change inside. Please, also rename the commit subjects to:

* compose: add AUDIT_CONTROL for rhel-8
* containers: add keycloak

And we are good to go. The failure in ubuntu rolling is unrelated, something we need to fix.

src/ansible/roles/keycloak/tasks/main.yml
```diff
@@ -0,0 +1,119 @@
+- name: Get Keycloak tags
+  uri:
+    url: https://api.github.com/repos/keycloak/keycloak/tags
```
pbrezina, line 3: Thinking of this, it may be safer to rely on latest release than on tags, ie `https://api.github.com/repos/keycloak/keycloak/releases/latest` and read the tag name from there - or read tarball_url directly if you don't require specific asset but can work with the full source code.

src/ansible/roles/keycloak/tasks/main.yml
spoore1: moved

src/ansible/roles/keycloak/tasks/main.yml
spoore1: I switched this to use latest but the Fedora 39 pipeline failed on this.   I'm rerunning now.

src/ansible/roles/keycloak/tasks/main.yml
```diff
@@ -0,0 +1,119 @@
...
+    export KC_HTTP_RELATIVE_PATH=/auth
+    /opt/keycloak/bin/kc.sh build
+    '''
+
+- name: Create Keycloak service env file
+  copy:
+    content: |
+      KEYCLOAK_ADMIN=admin
+      KEYCLOAK_ADMIN_PASSWORD={{ service.keycloak.admin_password }}
+      #KC_LOG_LEVEL=debug
```
pbrezina, line 82: Remove?

src/build.sh
```diff
@@ -8,14 +8,14 @@
 # |-----------------------------------------------------------------|
 # |                    base-ground                                  |
 # |-----------------------------------------------------------------|
```
pbrezina, line 10: Update as well?

src/ansible/roles/keycloak/tasks/main.yml
spoore1: ok, f39 worked so marking this resolved.

src/ansible/roles/keycloak/tasks/main.yml
spoore1: removed.

src/build.sh
spoore1: updated this banner

pbrezina (changes requested):
Two more places need to be changed:

- readme.md [there is chart at the end](https://github.com/SSSD/sssd-ci-containers#image-layers)
- push.sh so it is also pushed to quay

pbrezina (approved):

PR #38, opened 2023-04-19
ansible: move random.so build to standalone passkey role

pbrezina:
So it can be built on IPA as well in order to call `ipa
user-add-passkey --register` on IPA side.


ikerexxe (approved):
LGTM. Thanks!

PR #40, opened 2023-04-28
ansible: add firewalld

pbrezina:
We set the default zone to trusted so all incomming communication is
allowed. We can then just block specific ports in CI to simulate
unreachable servers with rich rules, for example:

```
firewall-cmd --add-rich-rule 'rule service name=ldap reject'
```


ikerexxe (approved):
LGTM. I was waiting for the CI issues to be fixed before approving

PR #42, opened 2023-05-17
Fix ubuntu rolling

jakub-vavra-cz:
Fix ubuntu rolling by changing uid of ci user and tweaking pip for installing flaky.


jakub-vavra-cz:
@pbrezina: This is ready for a review. Note that the red checks were just cancelled to save resources and get results of ubuntu-latest and ubuntu-rolling.

pbrezina:
Hi, thank you. I went with different approach to avoid conflicts with uid but keep the desired number as part of https://github.com/SSSD/sssd-ci-containers/pull/47.

PR #47, opened 2023-07-12
switch back to dnf-3 on rawhide

pbrezina:
DNF was rebased in Fedora 39 and its python bindings are not installed
in container by default. We need to install them in order to run the
dnf module.

https://fedoraproject.org/wiki/Changes/ReplaceDnfWithDnf5


pbrezina:
We must wait for https://github.com/ansible/ansible/pull/81141 to be released.

pbrezina:
I hit other issues when run against RC ansible version which includes the fix and there is discussion on fedora-devel to revert dnf5 in rawhide due to lots of issues. Let's revert to dnf-3 for now.

j-mracek:
I think there is incorrect expectation that package managed written in C++ that does not require Python will have installed in containers its python bindings. Containers are supposed to be as small as possible, right?

pbrezina:
> I think there is incorrect expectation that package managed written in C++ that does not require Python will have installed in containers its python bindings. Containers are supposed to be as small as possible, right?

Yes. On the other hand, with core tool like dnf, you should aim to provide maximum backward compatibility to avoid breaking scripts. But this can be discussed in the fesco ticket.

ikerexxe:
Two points before I approve. First, in the third commit you wrote `envirotnment` but it's `environment`. Second, `python3-flaky` installation doesn't seem to work, so I think you should take a look at it.

pbrezina:
Thank you, now all checks passed. I just fixed the typo in commit message.

ikerexxe (approved):
LGTM. Thanks for taking care of it!

PR #48, opened 2023-07-13
packages: update java for keycloak

pbrezina:
Java 11 support was removed from keycloak.

PR #49, opened 2023-07-13
kc java fix

spoore1:
Includes commits from PR #47 and #48 as well as a retry for downloading keycloak version information which was failing in gating pipelines.


src/ansible/roles/common/tasks/main.yml
```diff
@@ -64,6 +64,19 @@
     system: yes
     state: present
 
+- name: 'Check if user with uid {{ user.regular.uid }} already exists'
```
jakub-vavra-cz, line 67: Please do not do this, it is bound to break VMs in cloud as the account that is provisioned as default is the only one accessible with sudo priviledges.

jakub-vavra-cz (changes requested):
Please do not remove the accounts with default uid. The ci account should just use different uid instead.

src/ansible/roles/common/tasks/main.yml
pbrezina: This is for containers. As far as I know you are using different uid value on vms.

pbrezina: https://github.com/SSSD/sssd-ci-containers/pull/21/commits/5daa9e7e9375607be15668c92db16f4ea33f4f66#diff-39d9c8858e131ac0b5ac9afffe35f2a1c4f31f5a49536b15ae338c0660ca5671R12

So what happens is that it will make sure that uid 1024 is not present in the system before creating the user.

pbrezina: And since uid 1024 does not exist on the virtual machine, it is a noop.

pbrezina:
Scott, I have fixed all build issues with https://github.com/SSSD/sssd-ci-containers/pull/47 keycloak build is disabled. Please, revert the commit in this PR and fix issues with keycloak.

src/ansible/roles/common/tasks/main.yml
jakub-vavra-cz: I would rather see changing the ci user.regular.uid on collision so we can get rid of the override in first place.
Removing account to replace it with a new one with the same uid is generally considered as a bad practice.

At least please make sure that the ci user is not re-created so it can figure out that the user it found is the same one created by the playbook (on a re-run).

You closed my other MR that was touching/fixing the uids on ubuntu rolling completely ignoring the second fix that was addressing pip issue. https://github.com/SSSD/sssd-ci-containers/pull/42/files#diff-0fb17d023b75aba7561865659127745cb46523c950aee204c80b8601928561d6R200

pbrezina: > I would rather see changing the ci user.regular.uid on collision so we can get rid of the override in first place. Removing account to replace it with a new one with the same uid is generally considered as a bad practice.
> 
> At least please make sure that the ci user is not re-created so it can figure out that the user it found is the same one created by the playbook (on a re-run).
> 
> You closed my other MR that was touching/fixing the uids on ubuntu rolling completely ignoring the second fix that was addressing pip issue. https://github.com/SSSD/sssd-ci-containers/pull/42/files#diff-0fb17d023b75aba7561865659127745cb46523c950aee204c80b8601928561d6R200

This is fixed as well, by installing the package via apt.

jakub-vavra-cz: Well clearly it is not solved in this MR.
```
  - name: Install additional python packages
    pip:
      name:
      - flaky
      - python3-flaky
  when: "'client_devel' in group_names"
  ```

Comment (jakub-vavra-cz) [reply src/ansible/roles/common/tasks/main.yml:67]: Nevertheless the playbook will delete and re-create needlessly the ci user on rerun.
```
getent passwd {{ user.regular.uid }} | cut -d ':' -f 1 | grep -v '^{{ user.regular.name }}$'
```

Comment (pbrezina) [reply src/ansible/roles/common/tasks/main.yml:67]: > Well clearly it is not solved in this MR.
> 
> ```
>   - name: Install additional python packages
>     pip:
>       name:
>       - python3-flaky
>   when: "'client_devel' in group_names"
> ```

Because this PR just cherry-picked commits from my work.

Comment (pbrezina) [reply src/ansible/roles/common/tasks/main.yml:67]: This is the one: https://github.com/SSSD/sssd-ci-containers/pull/47 too bad you didn't comment there.

Comment (pbrezina) [reply src/ansible/roles/common/tasks/main.yml]: > Nevertheless the playbook will delete and re-create needlessly the ci user on rerun.
> 
> ```
> getent passwd {{ user.regular.uid }} | cut -d ':' -f 1 | grep -v '^{{ user.regular.name }}$'
> ```

https://github.com/SSSD/sssd-ci-containers/pull/50

**Comment by spoore1** [comment]
> Scott, I have fixed all build issues with #47 keycloak build is disabled. Please, revert the commit in this PR and fix issues with keycloak.

Ok, I'll take a look.  I started working before I saw this.  I've review #47 and take another look at the keycloak issue.

Should we review and merge #47 and then I can address keycloak separately here instead of me trying to cherry-pick everything needed into this one?

**Comment by pbrezina** [comment]
> > Scott, I have fixed all build issues with #47 keycloak build is disabled. Please, revert the commit in this PR and fix issues with keycloak.
> 
> Ok, I'll take a look. I started working before I saw this. I've review #47 and take another look at the keycloak issue.
> 
> Should we review and merge #47 and then I can address keycloak separately here instead of me trying to cherry-pick everything needed into this one?

It is already merged.

**Comment by spoore1** [comment]
> > > Scott, I have fixed all build issues with #47 keycloak build is disabled. Please, revert the commit in this PR and fix issues with keycloak.
> > 
> > 
> > Ok, I'll take a look. I started working before I saw this. I've review #47 and take another look at the keycloak issue.
> > Should we review and merge #47 and then I can address keycloak separately here instead of me trying to cherry-pick everything needed into this one?
> 
> It is already merged.

Oh even better.  Then I'll clean this up.   Thanks.

**Review by jakub-vavra-cz — APPROVED** [review]
The changes that need broader discussion were moved elsewhere. LGTM.

File: src/ansible/roles/keycloak/tasks/main.yml [hunk 2]
```diff
@@ -23,9 +30,10 @@
     create_home: yes
 
 - name: Download software zipfile
-  get_url:
-    url: "{{ keycloak_url }}"
-    dest: /tmp/keycloak.tgz
+  shell: curl -L {{ keycloak_url }} -o /tmp/keycloak.tgz
```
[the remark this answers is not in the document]
spoore1: switched this to command.

src/ansible/roles/packages/tasks/Debian.yml
```diff
@@ -87,16 +87,27 @@
...
       - unzip
       - ca-certificates
-      - ca-certificates-java
+      - openjdk-17-jre-headless
+  - name: Remove openjdk-11 workaround to ca-certificates-java dep issue
+    apt:
+      state: absent
+      update_cache: yes
+      name:
+      - openjdk-11-jre-headless
```
pbrezina, line 112: Since you install ca-certificates-jave first, Is this still needed or is it a leftover from debugging?

src/ansible/roles/keycloak/tasks/main.yml
```diff
@@ -1,15 +1,19 @@
+#- name: Get Keycloak latest version info
+#  uri:
+#    url: https://api.github.com/repos/keycloak/keycloak/releases/latest
+#    method: GET
+#    return_content: yes
+#    status_code: 200
+#    body_format: json
+#  register: result
```
pbrezina, line 8: I think you can just remove this comment as the change will be visible in git history.

src/ansible/roles/keycloak/tasks/main.yml
```diff
@@ -23,9 +27,10 @@
     create_home: yes
 
 - name: Download software zipfile
-  get_url:
-    url: "{{ keycloak_url }}"
-    dest: /tmp/keycloak.tgz
+  command: curl -L {{ keycloak_url }} -o /tmp/keycloak.tgz
+#  get_url:
+#    url: "{{ keycloak_url }}"
+#    dest: /tmp/keycloak.tgz
```
pbrezina, line 33: I think you can just remove this comment as the change will be visible in git history.

pbrezina (commented):
Thank you, see comments inline.

src/ansible/roles/packages/tasks/Debian.yml
spoore1: I think I was just being overly cautious and removing it in case it caused issues with Keycloak with both 11 and 17 installed.  I removed the code though and the pipelines passed so I'm dropping it.

src/ansible/roles/keycloak/tasks/main.yml
spoore1: done.

src/ansible/roles/keycloak/tasks/main.yml
spoore1: done

pbrezina (approved):

PR #50, opened 2023-07-18
Add yum to rawhide and don't remove ci user if it exist

pbrezina:


pbrezina:
Multihost tests and rawhide fails because yum is not available (we removed it together with dnf5) so we add it back now.

The second patch is per Jakub's request as he's adding idempotency requirement to provision VMs as well with the same playbook.

src/ansible/roles/common/tasks/main.yml
```diff
@@ -75,7 +75,7 @@
     state: absent
     remove: yes
     force: yes
-  when: regular_user_check.stdout | length > 0
+  when: regular_user_check.stdout | length > 0 and regular_user_check.stdout != user.regular.name
```
jakub-vavra-cz, line 78: Without checking I would be vary if there is not <LF> characters in  regular_user_check.stdout

pbrezina: Works in my tests.

jakub-vavra-cz (approved):
LGTM.

ikerexxe (approved):
LGTM. Thanks for the patch!

PR #51, opened 2023-07-20
provision: move random.so into standalone rpm

pbrezina:
To avoid pulling in build dependencies in minimal testing environment.


ikerexxe (approved):
I was also thinking about doing something similar, but my idea was to move the code to a separate repository and let COPR build it when the repository changed.

In any case, this is an improvement from the current workflow so I'm approving it.

ikerexxe:
> It is actually built in copr and it will be rebuilt on any change. https://copr.fedorainfracloud.org/coprs/g/sssd/ci-deps/monitor/

I had a different idea to reach to the same point :rofl:

pbrezina:
Rebased on top of Jakub's changes.

ikerexxe:
Pavel you'll need to rebase this PR due to https://github.com/SSSD/sssd-ci-containers/pull/54

pbrezina:
Rebased.

ikerexxe (approved):

jakub-vavra-cz (approved):
The ansible parts look good. I can't really comment on the rpm/copr side of things as I am not familiar enough with that.

PR #53, opened 2023-07-25
fix: Add defaut value for extended_packageset, fix renamed skip_addc_…

jakub-vavra-cz:
The variable extended_packageset did not have a default value resulting in failure when it was not in idmci metadata.
Issue did not manifest for containers as the variable is in the inventory.
Variable skip_addc_install was not renamed in all places.


pbrezina (approved):
It would be nice to split this into two commits, and prefix it with "provision" (area of code that you touch) but ack.

PR #55, opened 2023-08-29
Add ip address definition for ipa server installation.

jakub-vavra-cz:
When installing in openstack the server pulls link-local ipv6 address. This breaks the installation so we need to define the address implicitly to prevent that.

### DIFF
--- a/src/ansible/roles/ipa/tasks/main.yml
+++ b/src/ansible/roles/ipa/tasks/main.yml
@@ -1,3 +1,13 @@
+- name: Configure IPA IP address (from metadata)
+  set_fact:
+    ipa_ip: "{{ hostvars[inventory_hostname]['meta_ip'] }}"
+  when: '"meta_ip" in hostvars[inventory_hostname]'
+
+- name: Configure IPA IP address (from ssh)
+  set_fact:
+    ipa_ip: "{{ hostvars[inventory_hostname]['ansible_default_ipv4']['address'] }}"
+  when: '"meta_ip" not in hostvars[inventory_hostname]'
+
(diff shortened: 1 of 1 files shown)


pbrezina:
CI fails. Isn't it possible to just disable IPv6 stack in openstack?

jakub-vavra-cz:
Fixed the issue in ci. Looks like the code did not work with the available version of ansible.

src/ansible/roles/ipa/tasks/main.yml
```diff
@@ -1,3 +1,13 @@
+- name: Configure IPA IP address (from metadata)
+  set_fact:
+    ipa_ip: "{{ hostvars['master.ipa.test']['meta_ip'] }}"
+  when: '"meta_ip" in hostvars["master.ipa.test"]'
+
+- name: Configure IPA IP address (from ssh)
+  set_fact:
+    ipa_ip: "{{ hostvars[inventory_hostname]['ansible_default_ipv4']['address'] }}"
+  when: '"meta_ip" not in hostvars["master.ipa.test"]'
```
pbrezina, line 9: Shouldn't be `master.ipa.test` replaced with `inventory_hostname` everywhere?

jakub-vavra-cz: It probably could, will do that.

jakub-vavra-cz: Done

pbrezina:
Thanks, code-wise ack. Let's wait for the CI.

pbrezina (approved):

PR #56, opened 2023-08-31
provision: Add krb5.conf for realm join

jakub-vavra-cz:
We need krb5.conf  with rdns = false for realm join to work properly in environment where reverse dns records are different that internal names.

### DIFF
--- a/src/ansible/roles/client/tasks/main.yml
+++ b/src/ansible/roles/client/tasks/main.yml
@@ -55,6 +55,14 @@
       path:  /enrollment/samba.keytab
     register: enrollment_samba_1
 
+  - name: Create /etc/krb5.conf for samba join
+    template:
+      src: krb5.conf
+      dest: /etc/krb5.conf
+      owner: root
+      group: root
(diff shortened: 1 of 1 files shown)


pbrezina:
There is already "Create /etc/krb5.conf" online line 164. Wouldn't it be sufficient to just move this before  "Join IPA domain"?

jakub-vavra-cz:
The krb5.conf is cleaned after joining backends so just moving it would not work:
https://github.com/SSSD/sssd-ci-containers/blob/master/src/ansible/roles/client/tasks/main.yml#L42
https://github.com/SSSD/sssd-ci-containers/blob/master/src/ansible/roles/client/tasks/main.yml#L82
https://github.com/SSSD/sssd-ci-containers/blob/master/src/ansible/roles/client/tasks/main.yml#L129

The issue affected both joining AD and samba so this was the least invasive way to fix it.

pbrezina (approved):

pbrezina:
Ok, ack.

PR #60, opened 2023-09-21
build: add ipa-devel container with SSSD build dependencies

pbrezina:
So we are able to easily rebuilt SSSD on the IPA server to test changes to
SSSD's server mode.

Also add a patch to execute script on multiple containers, we'll use it to
build and install SSSD on client and ipa with one step.

---

We need to start rebuilding SSSD on the IPA server during PR CI to test changes to the server mode. The first commit ensures that build dependencies are preinstalled, the second commit is to avoid duplication in GH workflow.


ikerexxe (commented):
Out of ignorance, is there any difference between building sssd for client or server? Wouldn't it be better to build it once, and deploy it everywhere?

I also wonder, should we also update `readme.md` to include this new image?

pbrezina:
> Out of ignorance, is there any difference between building sssd for client or server? Wouldn't it be better to build it once, and deploy it everywhere?

There is no difference. The ultimate goal is to install SSSD from copr or packit builds instead of building it all over again. But we are not there yet. But this image is not only useful for PR CI but for local development as well so you don't have to install SSSD build dependencies again and again when you start up a container.

> I also wonder, should we also update `readme.md` to include this new image?

Done, thank you.

ikerexxe:
> There is no difference. The ultimate goal is to install SSSD from copr or packit builds instead of building it all over again. But we are not there yet. But this image is not only useful for PR CI but for local development as well so you don't have to install SSSD build dependencies again and again when you start up a container.

Thanks for the explanation. LGTM!

ikerexxe (approved):

PR #62, opened 2023-10-05
packages: install traffic control dependency

ikerexxe:


pbrezina:
LGTM, please check availability of iproute-tc in rhel8 and rhel9.

ikerexxe:
The package is available in both versions

src/ansible/roles/packages/tasks/Fedora.yml
```diff
@@ -55,6 +55,8 @@
       - firewalld
       - gdb
       - gdb-gdbserver
+      - iproute
+      - iproute-tc
```
pbrezina, line 59: This has to go to minimal package set since it is test dependency. So it needs to go to `Install systemd and minimal set of packages`.

We could also add it only to `Install test dependencies on client` but this would make it available only on client host. But I think this tool may be handy on other hosts as well.

pbrezina (changes requested):

pbrezina (approved):

PR #64, opened 2023-10-18
build: disable nscd if present

sumit-bose:
On some platforms nscd might be pulled in and started via dependencies.
Since running SSSD together with nscd is not supported it should be
disabled if present.

### DIFF
--- a/src/ansible/playbook_image_service.yml
+++ b/src/ansible/playbook_image_service.yml
@@ -4,6 +4,7 @@
   roles:
   - facts
   - firewall
+  - no_nscd
 
 - hosts: master.ldap.test
   gather_facts: no
--- a/src/ansible/roles/no_nscd/tasks/main.yml
+++ b/src/ansible/roles/no_nscd/tasks/main.yml
@@ -1,0 +1,17 @@
+- name: Populate service facts
+  service_facts:
+
+- name: Disable nscd.service
+  service:
+    name: nscd.service
+    enabled: no
+    state: stopped
+  when: "ansible_facts.services['nscd.service'] is defined"
+
+- name: Disable nscd.socket
+  service:
+    name: nscd.socket
+    enabled: no
+    state: stopped
+  when: "ansible_facts.services['nscd.socket'] is defined"
+
--- a/src/ansible/roles/packages/tasks/Debian.yml
+++ b/src/ansible/roles/packages/tasks/Debian.yml
@@ -57,8 +57,10 @@
       - adcli
       - freeipa-client
       - nfs-common
+      - nslcd
       - packagekit
       - realmd
+      - slapd
       - sssd
       - sssd-*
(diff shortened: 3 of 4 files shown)


pbrezina:
Thank you. LGTM. Can you move it to `common` please? That seems to be a better place.

src/ansible/roles/packages/tasks/main.yml
```diff
@@ -5,6 +5,23 @@
   with_first_found:
   - files: '{{ ansible_distribution | distro_includes(ansible_distribution_major_version) }}'
 
+- name: Populate service facts
+  ansible.builtin.service_facts:
```
pbrezina, line 9: Please, remove `ansible.builtin.` prefix, we don't have it on other places.

sumit-bose:
> Thank you. LGTM. Can you move it to `common` please? That seems to be a better place.

Hi,

I think `common` is run too early. I guess `nscd` is install as a dependency during `Install SSSD build and integration tests dependencies` and has to be disabled afterwards. Do you have any other suggestions where to put it?

bye,
Sumit

pbrezina:
I see. Maybe we can add new role here: https://github.com/SSSD/sssd-ci-containers/blob/master/src/ansible/playbook_image_service.yml#L6

sumit-bose:
> I see. Maybe we can add new role here: https://github.com/SSSD/sssd-ci-containers/blob/master/src/ansible/playbook_image_service.yml#L6
> 
> This will make sure it is called only once and after all packages are installed.

Hi,

this is too early as well. `Install SSSD build and integration tests dependencies` is one of the very last steps of building `client-devel` and as long as the packages are not installed the service cannot be disabled.

bye,
Sumit

pbrezina:
Ah, I misunderstood it. I suppose it is pulled by nslcd package and it is installed in a wrong place for Debian and Ubuntu.

It should be in `Install packages for client base image` since it is a requirement to run multihost/system tests. Fedora has in (nss-pam-ldapd) on correct place. https://github.com/SSSD/sssd-ci-containers/blob/master/src/ansible/roles/packages/tasks/Fedora.yml#L121 Jakub shifted some packages but only for Fedora, we should also do the same for Debian and Ubuntu. For now, it will be enough to just move the nslcd package.

`Install additional packages for client development image` is about installing development packages, needed to build SSSD and run integration tests, but not needed for the multihost/system tests.

sumit-bose:
Hi,

ok after moving `nslcd` and `slapd` to `Install packages for client base image` it was working with a new role. Do you prefer the new role or should it be better in `common` ?

bye,
Sumit

src/ansible/roles/packages/tasks/main.yml
sumit-bose: Fixed in the latest version.

pbrezina (approved):

pbrezina:
Thank you, ack.

PR #66, opened 2023-10-31
packages: add augeas

danlavu:


ikerexxe (commented):
By reading this PR I have some general doubts about your objective:
* Where do you need to install this dependency? Client, or all images?
* Is this an integration test dependency?
* Did you forget Debian, or is this package installed by default in that distribution?

danlavu:
> By reading this PR I have some general doubts about your objective:
> 
> * Where do you need to install this dependency? Client, or all images?
Client only
> * Is this an integration test dependency?
Yes, for authselect. 
> * Did you forget Debian, or is this package installed by default in that distribution?
I did forget, updated.

ikerexxe:
> * Is this an integration test dependency?
>   Yes, for authselect.

The answer seems contradictory, at least to my knowledge. Let's set a common ground and check the question again.

Integration testing is a type of software testing where two or more modules of an application are logically grouped together and tested as a whole. System testing is types of testing where tester evaluates the whole system against the specified requirements.

IIUC, you'd like to test authselect as a whole against the specified requirements. In that case, you'd like to do system testing.

Source:
https://www.softwaretestinghelp.com/types-of-software-testing/

src/ansible/roles/packages/tasks/Fedora.yml
```diff
@@ -206,6 +206,7 @@
     dnf:
       state: present
       name:
+      - augeas
```
ikerexxe, line 209: This should go in `Install test dependencies on client` task

src/ansible/roles/packages/tasks/Debian.yml
```diff
@@ -55,6 +55,7 @@
       update_cache: yes
       name:
       - adcli
+      - augeas-tools
```
ikerexxe, line 58: Create a new task called `Install test dependencies on client` after this task and add this dependency there.

src/ansible/roles/packages/tasks/Ubuntu.yml
```diff
@@ -45,6 +45,7 @@
       update_cache: yes
       name:
       - adcli
+      - augeas-tools
```
ikerexxe, line 48: Create a new task called `Install test dependencies on client` after this task and add this dependency there.

ikerexxe (commented):
Some minor comments and I think we'll be done

danlavu:
Ack, all changes are done, yes, it is functional testing, not an integration test. Makes sense to create a new section.  ;)

ikerexxe (approved):
LGTM

pbrezina:
Can you please change the commit message to `packages: add augeas`?

PR #67, opened 2023-11-05
adding backup dc to the root domain

danlavu:
Downstream tests will need a secondary domain controller for several test scenarios before they can be ported to the new framework.

- adding a new host to vagrant, ad1
- adding a new host to the inventory file
- creating new variable "join_domain" for when conditions
- updating dns files with host entry


pbrezina:
@jakub-vavra-cz Hi Jakub, you have some plans on integrating more complex AD environments into this project. Does this follow the plan?

PR #69, opened 2023-11-20
ci: Allow overriding trust

jakub-vavra-cz:
Allow overriding trust between IPA and samba and IPA and AD.

### DIFF
--- a/src/ansible/playbook_vm.yml
+++ b/src/ansible/playbook_vm.yml
@@ -23,6 +23,8 @@
     join_ldap: "{{ override_join_ldap | default('yes') | bool }}"
     join_samba: "{{ override_join_samba | default('yes') | bool }}"
     join_ipa: "{{ override_join_ipa | default('yes') | bool }}"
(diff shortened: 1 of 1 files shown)


spoore1 (commented):
I like having separate vars for trust vs join.  Is there any downside to defaulting to yes here though?

jakub-vavra-cz:
> I like having separate vars for trust vs join. Is there any downside to defaulting to yes here though?

Not any downside I am aware of. What could get some improvement is auto-detecting availability of AD/samba/IPA/ldap/kcm in the inventory so join_X and trust_x_y would not need overriding to "no" when the machine is missing.

pbrezina (approved):

pbrezina:
> > I like having separate vars for trust vs join. Is there any downside to defaulting to yes here though?
> 
> Not any downside I am aware of. What could get some improvement is auto-detecting availability of AD/samba/IPA/ldap/kcm in the inventory so join_X and trust_x_y would not need overriding to "no" when the machine is missing.

We could write some python extension for this kind of stuff.

PR #71, opened 2024-01-05
Fedora: remove 'nodocs' from dnf.conf::tsflags

alexey-tikhonov:
To allow man pages installation. This is to enable proper feature detection that rely on man pages content.

### DIFF
--- a/src/ansible/roles/packages/tasks/Fedora.yml
+++ b/src/ansible/roles/packages/tasks/Fedora.yml
@@ -1,3 +1,17 @@
+- name: Enable docs installation by default
+  block:
+  - name: Check dnf.conf exists
+    stat:
+      path: /etc/dnf/dnf.conf
+    register: dnf_stat_result
+
+  - name: Patch dnf.conf
+    replace:
+      path: /etc/dnf/dnf.conf
+      regexp: 'nodocs'
+      replace: ''
+    when: dnf_stat_result.stat.exists
+
 - name: Select packageset
   set_fact:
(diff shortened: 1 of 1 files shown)


ikerexxe:
I guess this problem only happens in Fedora, right?

Out of curiosity, which feature isn't correctly detected?

alexey-tikhonov:
> I guess this problem only happens in Fedora, right?

Looks so (in https://github.com/SSSD/sssd/pull/7049)


> Out of curiosity, which feature isn't correctly detected?

"--with-sssd-user=sssd" that relies on existence of "user" config option description in the man page.

ikerexxe (approved):
LGTM!

src/ansible/roles/packages/tasks/Fedora.yml
```diff
@@ -1,3 +1,9 @@
+- name: Patch dnf.conf
+  ansible.builtin.replace:
```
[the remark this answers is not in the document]
alexey-tikhonov: Done.

src/ansible/roles/packages/tasks/Fedora.yml
```diff
@@ -1,3 +1,17 @@
+- name: Fix dnf.conf
```
pbrezina, line 1: Enable docs installation by default?

alexey-tikhonov: Done.

pbrezina (approved):

PR #73, opened 2024-01-14
Update readme.md with up-passkey make target

mglantz:
Information was missing about the up-passkey make target which enables testing of passkey.


ikerexxe (approved):
LGTM!

PR #74, opened 2024-01-17
keys: Each key must be of a different type

aplopez:
A typo made all the ssh keys of type ecdsa when they must be of different types.


pbrezina:
Please,  you also need to regenerate the keys (by calling the script) and commit them (only the missing type, leave ecdsa intact).

aplopez:
> Please, you also need to regenerate the keys (by calling the script) and commit them (only the missing type, leave ecdsa intact).

Done

pbrezina (approved):

PR #77, opened 2024-02-25
Fixes and improvements for package installation during image builds

dpward:
These changes make image building more consistent between different distributions, by fixing discrepancies and re-using task lists where appropriate. (The task lists for RHEL now include those for CentOS; and the task lists for Ubuntu now include those for Debian.) This is intended to simply maintenance and allow for easier testing with other base images.

This also improves the build speed. Temporary volumes are now created and mounted at the paths where repository metadata is stored, so that it does not need to be re-downloaded for each image that is built (ipa, ldap, etc.) For Debian/Ubuntu, this metadata is now re-used across different package management operations for the same image.

Builds were performed successfully from these base images:

- registry.fedoraproject.org/fedora:{rawhide,40,latest}
- quay.io/centos/centos:{stream9,stream8}
- docker.io/library/debian:{unstable,testing,stable,oldstable}
- docker.io/library/ubuntu:{devel,rolling,latest}


pbrezina:
Wow, that's a lot of patches thank you. I'll let PR CI run and then review it if there is no major issue.

pbrezina:
Hmm, PR CI errors are not related to this pull request. It seems that some update started to cause issues when commiting to containers in the github runners. I need to look into it first.

pbrezina:
Hopefully, I fixed all PR CI issues, can you please rebase this pull request?

src/ansible/roles/packages/tasks/Fedora.yml
```diff
@@ -1,23 +1,11 @@
...
-
-  - name: Patch dnf.conf
+  - name: Enable docs installation by default
     replace:
       path: /etc/dnf/dnf.conf
       regexp: 'nodocs'
       replace: ''
-    when: dnf_stat_result.stat.exists
 
-- name: Select packageset
```
jakub-vavra-cz, line 15: For testing we try to reduce the set of installed packages this is needed for
"Install extended set of packages"

jakub-vavra-cz (changes requested):
Removal of the groups without base_ prefix from when clauses will break downstream automation that depends on the simple names.
We are not using two inventories for installing packages and configuring services but a single one that has only the names like nfs, client...
We are installing and configuring  VMs so this split to layers does not make any sense there and makes the configuration needlessly complex and prone to issues.

src/ansible/roles/packages/tasks/Fedora.yml
```diff
@@ -149,15 +134,18 @@
       name:
       - acl
       - 389-ds-base
-  when: "'base_ldap' in group_names or 'ldap' in group_names or 'ipa' in group_names"
```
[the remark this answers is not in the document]
jakub-vavra-cz: I am probably missing something, why would it be run twice and slow down the build?
When containers are built the phases for base with packages base_xxx are separate from the services so packages role is not executed again in services phase.
The inventory for VMs is already huge and complicated and we do not want it to be inflated by the extra groups like base_XXX.
I would rather see the whole base_XXX deal dropped and used just roles like ipa, ldap ./.. instead.

src/ansible/roles/facts/tasks/main.yml
```diff
@@ -1,6 +1,2 @@
-- name: 'Include distribution specific tasks [{{ ansible_distribution }} {{ ansible_distribution_major_version }}]'
-  include_tasks: '{{ include_file }}'
-  loop_control:
-    loop_var: include_file
-  with_first_found:
-  - files: '{{ ansible_distribution | distro_includes(ansible_distribution_major_version) }}'
+- name: 'Include distribution specific tasks [{{ ansible_distribution }}]'
```
pbrezina, line 1: Please, keep the plugin. It may not be needed now, but it most certainly will be needed in the future.

src/build.sh
```diff
@@ -156,3 +172,5 @@ compose down
 # Create development images with additional packages
 build_base_image "ci-client:${TAG}" client-devel
 build_base_image "ci-ipa:${TAG}" ipa-devel
+
+rm --recursive --force "$SHARED_CACHE_ROOT"
```
pbrezina, line 175: This should go to cleanup?

pbrezina (changes requested):
Hi, thank you for the changes. Please, keep the group names as is at this moment. AFAIK the tasks are not run twice and it is needed so far. But it is certainly something we want to consolidate in the future.

andreboscatto:
Hi @dpward 

It has been a while since we left a few comments in this PR, will you be able to review those? If there is anything we can do to help, please let us know.

Kind regards

PR #78, opened 2024-02-27
Replace buildroot repo with crb/powertools

alexey-tikhonov:


alexey-tikhonov:
centos-8: `No package umockdev available`...

@martinpitt, hi.
What repo provides `umockdev` on centos-9?

alexey-tikhonov:
I guess this is blocked until either 'umockdev' gets into regular repo (maybe EPEL?) or 'passkey' tests are re-done to use software token.

martinpitt:
@alexey-tikhonov TBH I have no idea how this works. https://mirror.stream.centos.org/9-stream/CRB/x86_64/os/Packages/ indeed doesn't have it, I'm not sure where RHEL buildroot packages end up in CentOS -- it does get released to and imported from CentOS, but apparently not published? :cry: 

This *will* not go into RHEL 9 AppStream -- it is deliberately a buildroot-only package. So I suppose the bug is that CentOS stream doesn't publish these?

I can easily provide a C9S build in a COPR. No idea about EPEL (it must only contain packages which are not in RHEL, but does buildroot count or not?) but that may be doable as well.

alexey-tikhonov:
> @alexey-tikhonov TBH I have no idea how this works. https://mirror.stream.centos.org/9-stream/CRB/x86_64/os/Packages/ indeed doesn't have it, I'm not sure where RHEL buildroot packages end up in CentOS -- it does get released to and imported from CentOS, but apparently not published? 😢
> 
> This _will_ not go into RHEL 9 AppStream -- it is deliberately a buildroot-only package. So I suppose the bug is that CentOS stream doesn't publish these?

Then I think it's not a bug that CentOS stream repos doesn't have it either.

Did you consider shipping it in CRB repo?

> I can easily provide a C9S build in a COPR. No idea about EPEL (it must only contain packages which are not in RHEL, but does buildroot count or not?) but that may be doable as well.

I think buildroot doesn't "count" but that's probably not worth the effort. We can merely keep enabling centos stream buildroot in our CI images...

martinpitt:
> Did you consider shipping it in CRB repo?

I would, but how do I do that? I had thought about this as "the dumping ground of all the packages that are built but not published in BaseOS/AppStream, but apparently that's not the case. It gets [built in CentOS Stream](https://kojihub.stream.rdu2.redhat.com/koji/buildinfo?buildID=35345), it just doesn't seem to end up anywhere.

martinpitt:
>  ` - {name: 'buildroot', url: 'https://kojihub.stream.centos.org/kojifiles/repos/c8s-build/latest/$basearch'}`

Ooh -- so there *is* a buildroot for CentOS stream -- what' wrong with this? (Sorry, I should have looked at the patch first). There is no commit message here with a rationale why you want to do this replacement.

alexey-tikhonov:
> Ooh -- so there _is_ a buildroot for CentOS stream -- what' wrong with this?

@pbrezina, do you remember rationale why did we want to get rid of the need for buildroot?

pbrezina:
There were requests by customers to be able to build SSSD in RHEL/CentOS, there were very few packages available only in buildroot so we moved them to standard repository. SSSD now builds without buildroot, but it is perfectly fine to keep it here for non-build dependencies if needed. Or we can build the required packages in copr sssd-ci-deps project.

alexey-tikhonov:
Ok, thanks.

PR #85, opened 2024-03-20
roles: disable expiration for administrator in samba

danlavu:


danlavu:
Samba hasn't worked out of the gate without disabling the expiration for the administrator user.

src/ansible/roles/samba/tasks/main.yml
```diff
@@ -183,3 +183,7 @@
     name: samba-sysvolreset.service
     enabled: yes
     state: started
+
+- name: Disable account expiration for Administrator
+  shell: |
```
jakub-vavra-cz, line 188: Could be probably command: instead.

jakub-vavra-cz (approved):
LGTM

PR #86, opened 2024-03-27
Add stream-10-devel container

alexey-tikhonov:


alexey-tikhonov:
`dnf copr enable -y @sssd/ci-deps` fails because copr doesn't have chroot for stream-10 yet

src/ansible/roles/packages/tasks/Fedora.yml
```diff
@@ -231,24 +228,33 @@
...
+      state: present
+      name:
+      - libfaketime
+      - lcov
+      - nss-pam-ldapd
+      - openldap-servers
+      - python3-pycodestyle
+      - tig
+      - tmate
+    when: passkey_support
```
alexey-tikhonov, line 256: This is just a quick test to see of anything else is missing.

PR #88, opened 2024-04-15
test

pbrezina:


pbrezina:
I have identified the root cause for c8s build failure, its `jaxb-fi` package from buildroot. I reported this to centos people so hopefully, it gets fixed soon.

PR #89, opened 2024-04-16
Add stream-10-devel container

alexey-tikhonov:


alexey-tikhonov:
@pbrezina , https://copr.fedorainfracloud.org/coprs/g/sssd/ci-deps/ got 'Centos-stream 10' but not 'EPEL 10'.
Why / how to enable EPEL10?

pbrezina:
Copr does not have epel10 chroot available at this moment, we will need to wait.

pbrezina:
Or we can switch to centos10 when running dnf copr enable...

.github/workflows/build.yml
```diff
@@ -78,6 +78,7 @@ jobs:
         image: [
           { base: 'quay.io/centos/centos:stream8', tag: 'centos-8', extra: '', unavailable: 'samba' },
           { base: 'quay.io/centos/centos:stream9', tag: 'centos-9', extra: 'centos-latest', unavailable: 'samba' },
```
pbrezina, line 80: remove centos-latest

alexey-tikhonov:
> Or we can switch to centos10 when running dnf copr enable so we don't depend on epel.

In centos10 playbooks only?

.github/workflows/build.yml
alexey-tikhonov: Removed.

alexey-tikhonov:
`- No package tmate available.`

alexey-tikhonov:
hm... https://copr.fedorainfracloud.org/coprs/g/sssd/ci-deps/packages/ :: 'Last build status' is pretty "red"...

pbrezina:
Looks like there is issue with ppc64 - that will resolve on its own, I assume. tmate fails on all centos10 with `No matching package to install: 'msgpack-devel >= 1.1.8'` which is weird, since msgpack was built successfully on all but ppc64.

lcov was used for intg tests, I think we can drop it now.

tmate is useful when you want to share issues with team mate, but it is not needed for CI. I will look into it tomorrow and see if we can make it work. If not, we can drop it.

pbrezina:
I rebuilt packages in order and the only one still failing is libfaketime, I don't remember what we use it for though. That one simply exits with

```
+ FAKETIME_COMPILE_CFLAGS=BOGUS
+ '[' BOGUS == BOGUS ']'
+ echo 'SHOULD NEVER REACH HERE ... YOU HAVE AN UNTESTED VERSION+ARCH, see rpm spec for details ... ABORT'
+ exit 1
error: Bad exit status from /var/tmp/rpm-tmp.1rnZcM (%build)
    %patchN is deprecated (3 usages found), use %patch N (or %patch -P N)
    Bad exit status from /var/tmp/rpm-tmp.1rnZcM (%build)
```

pbrezina:
Btw, you can disable other distros for testing purpose to make pr ci faster. See: https://github.com/SSSD/sssd-ci-containers/pull/91

alexey-tikhonov:
'virtualsmartcard' also fails : https://download.copr.fedorainfracloud.org/results/@sssd/ci-deps/centos-stream-10-x86_64/07325915-virtualsmartcard/builder-live.log.gz

alexey-tikhonov:
It's also unclear where 'vpcd' - https://github.com/SSSD/sssd-ci-containers/blob/master/src/ansible/roles/packages/tasks/Fedora.yml#L170 - should be installed from.

alexey-tikhonov:
`- No package mock available.`

alexey-tikhonov:
`No package fakeroot available`
It's missing in https://kojihub.stream.centos.org/kojifiles/repos/c10s-build/latest/x86_64/pkglist

.github/workflows/build.yml
pbrezina: I meant on the line 80 (from centos-9, since you added the tag to centos-10). But I guess both variants work since 10 is still in early development.

pbrezina (approved):

PR #93, opened 2024-04-26
packages: Use centos-stream-10 chroot for c10s and rhel-10

jakub-vavra-cz:
There are chroots in copr from rhel-10 or epel-10 yet so we will use c10s one for now.


alexey-tikhonov:
I tested that it works on RHEL10 VM in the sense repo can be enabled and packages installed.

We can get into all kinds of different troubles if C10S and RHEL10 use different build settings. But we can try.

alexey-tikhonov (approved):

PR #94, opened 2024-05-01
Install sssd-ldap package on IPA container

liswang89:
This PR is to resolve the issue encountered during the conversion of test_ldapapi to the new test framework.

In this feature, IPA, which stands for Identity Policy and Audit, serves as the upstream name. 389 operates as the backend, functioning as an LDAP server. To establish a connection to the LDAP database over a UNIX socket, it is necessary to install the sssd.ldap package on the IPA host.

### DIFF
--- a/src/ansible/roles/packages/tasks/Debian.yml
+++ b/src/ansible/roles/packages/tasks/Debian.yml
@@ -72,6 +72,13 @@
       - augeas-tools
   when: "'base_client' in group_names or 'client' in group_names"
 
+  - name: Install test dependencies on IPA server
+    apt:
+      state: present
+      name:
+      - sssd-ldap*
+  when: "'base_ipa' in group_names or 'ipa' in group_names"
+
 - name: Install packages for NFS base image
   block:
   - name: Install NFS
--- a/src/ansible/roles/packages/tasks/Debian.yml
+++ b/src/ansible/roles/packages/tasks/Debian.yml
@@ -70,14 +70,8 @@
       update_cache: yes
       name:
       - augeas-tools
+      - 389-ds-base
   when: "'base_client' in group_names or 'client' in group_names"
-
-  - name: Install test dependencies on IPA server
-    apt:
-      state: present
-      name:
-      - sssd-ldap*
-  when: "'base_ipa' in group_names or 'ipa' in group_names"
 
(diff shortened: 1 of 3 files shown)


danlavu:
The 389 sssd client isn't configured on the ldap container, so we decided to use the ipa container for testing instead. Lisa is working on a test case automation, multihost/basic/test_ldapapi.py  where sssd connects using ldapi protocol (UNIX socket) instead of TCP. So this needs to be executed on the server, with sssd configured using the ldap provider, that is why the package needs to be installed. 

@liswang89 , I do think we should create another task to install client packages on the IPA server, since the task is called install IPA. You also want to check Ubuntu and Debian yml files because they will require the package to be installed as well. The package names will likely not be the same, please confirm the package name for the proper distros.

danlavu (changes requested):
Please create a new task similar to this one but name it "Install test dependencies on IPA server"

And do the following for Ubuntu and Debian, please.

liswang89:
@danlavu Thank you for your comments. A task was created to address the issue, and the PR has been updated accordingly.

https://issues.redhat.com/projects/SSSD/issues/SSSD-7358

jakub-vavra-cz:
Hi, didn't test_ldapapi use local ldap on the client? Wouldn't it make more sense to run this in ldap container/vm instead?

danlavu:
> Hi, didn't test_ldapapi use local ldap on the client? Wouldn't it make more sense to run this in ldap container/vm instead?

That means we'd have to install 389 on the client. LDAPI connects over a UNIX socket, so it must be tested on the same host as 389.

jakub-vavra-cz:
> > Hi, didn't test_ldapapi use local ldap on the client? Wouldn't it make more sense to run this in ldap container/vm instead?
> 
> That means we'd have to install 389 on the client. LDAPI connects over a UNIX socket, so it must be tested on the same host as 389.

LDAPI connects to ldap not to IPA that's why I am asking why not use ldap container instead.

liswang89:
@jakub-vavra-cz Since the client is not configured for LDAP, it would be easier to install the dependencies from IPA.

pbrezina:
We can also install 389ds on the client container, no problem there. It would require more simple sssd configuration then using ldap provider with ipa.

PR #95, opened 2024-05-06
provision: do not join domains automatically in Idm CI

pbrezina:
Currently, we provide machines that are already fully setup with the
    client enrolled into all domains and trusts established. This is no
    longer necessary. With this patch, we will enroll into the domains
    using per-topology setup if needed.

    This applies only to downstream CI. Containers are still fully provisioned.

Please note that this needs to be closed together with changes to other repositories:
* https://github.com/SSSD/sssd/pull/7349
* https://github.com/next-actions/pytest-mh/pull/54
* https://github.com/next-actions/pytest-mh/pull/54

### DIFF
--- a/src/ansible/playbook_vm.yml
+++ b/src/ansible/playbook_vm.yml
@@ -19,12 +19,12 @@
     passkey_support: "{{ override_passkey_support | default('no') | bool }}"
     user_regular_uid: 1024
     ansible_become: yes
-    join_ad: "{{ override_join_ad | default('yes') | bool }}"
-    join_ldap: "{{ override_join_ldap | default('yes') | bool }}"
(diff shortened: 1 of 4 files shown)


danlavu (approved):
LG2M

justin-stephenson (approved):
Ack.

pbrezina:
> We should change also https://github.com/SSSD/sssd-ci-containers/blob/master/src/ansible/group_vars/all so containers are joining to backends only when explicitly requested as the only use-case is for manual testing. Both downstream and upstream ci should use the framework topology fixtures to join to reduce discrepancies and use the same code path. Otherwise we risk additional set of issues missed in prci.

This will come later as it will require more changes to keep development use case functional.

PR #98, opened 2024-06-03
build: use dnf5 when available

pbrezina:
dnf5 is now a protected package in rawhide and cannot be removed.
We need to start using it.


pbrezina:
Closing in favor of https://github.com/SSSD/sssd-ci-containers/pull/99

PR #100, opened 2024-06-06
ci: Switch c8s repos to vault

jakub-vavra-cz:
- The original centos mirrors no longer work so we need to switch to vault repos instead.
- ansible-core-2.17 is not compatible with CentOS 8 so we need to use older 2.16 version.
- C8s Buildroot is not available atm. in koji so we in non-mandatory repo.
- Crb (powertools) repo is needed for some packages from client-devel
- We need to make sure seliinux-poilicy-targeted needed by smartcards is installed on c8s.

### DIFF
--- a/.github/actions/install-dependencies/action.yml
+++ b/.github/actions/install-dependencies/action.yml
@@ -1,12 +1,30 @@
 name: 'Install dependencies'
 description: 'Install dependencies for building and publishing containers'
+
+inputs:
+  tag:
+    description: Built container tag.
+    required: false
+    default: ""
+
 runs:
   using: "composite"
   steps:
-  - name: Install packages
+  - name: Install python packages with older ansible for Centos 8
+    if: ${{ inputs.tag == 'centos-8' }}
+    shell: bash
+    run: |
+      sudo pip3 install ansible==9.8 passlib
+
+  - name: Install python packages
+    if: ${{ inputs.tag != 'centos-8' }}
     shell: bash
     run: |
       sudo pip3 install ansible passlib
+
+  - name: Install deb packages
+    shell: bash
+    run: |
       sudo apt-get update
       sudo apt-get install -y podman docker-compose
 
--- a/src/build.sh
+++ b/src/build.sh
@@ -57,6 +57,12 @@ function base_exec {
   ${DOCKER} exec sssd-wip-base /bin/bash -c "$1"
 }
 
+function c8s_repo {
+    # Update repos to working ones
+    ${DOCKER} exec sssd-wip-base /bin/bash -c 'grep -q "CentOS Stream 8" /etc/os-release && sed -i "s/mirrorlist/#mirrorlist/g" /etc/yum.repos.d/CentOS-* || true'
+    ${DOCKER} exec sssd-wip-base /bin/bash -c 'grep -q "CentOS Stream 8" /etc/os-release && sed -i "s|#baseurl=http://mirror.centos.org|baseurl=http://vault.centos.org|g" /etc/yum.repos.d/CentOS-* || true'
+}
+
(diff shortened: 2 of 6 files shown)


src/build.sh
```diff
@@ -130,7 +136,7 @@ fi
 
 # Create services
 compose up --detach
-ansible-playbook $ANSIBLE_OPTS ./ansible/playbook_image_service.yml
+ansible-playbook -v $ANSIBLE_OPTS ./ansible/playbook_image_service.yml
```
pbrezina, line 139: Remove -v, there is ANSIBLE_OPTS env var if you need to set it.

src/build.sh
```diff
@@ -94,10 +100,10 @@ function build_base_image {
   echo "Building $name from $from"
   ${DOCKER} run --security-opt seccomp=unconfined --name sssd-wip-base --detach -i "$from"
   if [ $name == 'base-ground' ]; then
+    c8s_repo
     base_install_python
   fi
-
-  ansible-playbook --limit "`echo $name | sed -r 's/-/_/g'`" ./ansible/playbook_image_base.yml
+  ansible-playbook -v --limit "`echo $name | sed -r 's/-/_/g'`" ./ansible/playbook_image_base.yml
```
pbrezina, line 106: Remove -v.

.github/actions/install-dependencies/action.yml
```diff
@@ -1,12 +1,31 @@
...
+      sudo pip3 install ansible==9.8 passlib
+
+  - name: Install python packages
+    if: ${{ inputs.tag != 'centos-8' }}
     shell: bash
     run: |
       sudo pip3 install ansible passlib
+
+  - name: Install deb packages
+    if: ${{ inputs.tag != 'centos-8' }}
```
pbrezina, line 26: Why? podman and docker-compose must be installed.

pbrezina (changes requested):
Hi, please, see minor comments inside.

Please, add thorough description of why the change is needed in each commit message. `ci` prefix is also incorrect since you are touching ci only in one commit and as a result of change in actions. You can you `provision` or `centos-8` instead.

pbrezina:
> * The original centos mirrors no longer work so we need to switch to vault repos instead.
> 
>     * ansible-core-2.17 is not compatible with CentOS 8 so we need to use older 2.16 version.
> 
>     * C8s Buildroot is not available atm. in koji so we in non-mandatory repo.
> 
>     * Crb (powertools) repo is needed for some packages from client-devel
> 
>     * We need to make sure seliinux-poilicy-targeted needed by smartcards is installed on c8s.

This should go into the relevant commit messages so it is clear why you did this change from commit history.

.github/actions/install-dependencies/action.yml
jakub-vavra-cz: fixed

src/build.sh
jakub-vavra-cz: done

src/build.sh
jakub-vavra-cz: done

jakub-vavra-cz:
> > * The original centos mirrors no longer work so we need to switch to vault repos instead.
> >   
> >   * ansible-core-2.17 is not compatible with CentOS 8 so we need to use older 2.16 version.
> >   * C8s Buildroot is not available atm. in koji so we in non-mandatory repo.
> >   * Crb (powertools) repo is needed for some packages from client-devel
> >   * We need to make sure seliinux-poilicy-targeted needed by smartcards is installed on c8s.
> 
> This should go into the relevant commit messages so it is clear why you did this change from commit history.

Reworded the commit messages.

pbrezina (approved):

PR #103, opened 2024-06-11
Package: Add sssd-kcm explicitly

madhuriupadhye:
While installing sssd and dependancy add sssd-kcm
explicitly as its not downloading with 'sssd-*'


jakub-vavra-cz (approved):
LGTM

PR #104, opened 2024-06-20
Explicitly allow all capabilities required by SSSD in client

alexey-tikhonov:
and IPA containers.
Currently only 'CAP_DAC_READ_SEARCH' was missing, so probably other were added by default. But it is more clear to list everything explicitly.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -32,6 +32,11 @@ services:
     - AUDIT_CONTROL
     - SYS_CHROOT
     - NET_ADMIN
+    - CAP_CHOWN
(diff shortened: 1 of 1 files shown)


alexey-tikhonov:
It would be better to run containers in user-ns, so that granting any capabilities in host ns would be needed, but I have no idea how to do this with 'docker-compose'.

sumit-bose (approved):
Hi,

thanks for the fix, now https://github.com/SSSD/sssd/pull/7445 is working for me on the SSSD-CI client. ACK.

bye,
Sumit

PR #105, opened 2024-07-22
Match roles against host groups, tweak ssh server.

jakub-vavra-cz:
Run roles against host groups instead of hostnames as a preparation for extending the topologies.
Make sure ssh is not broken for ad-hoc added machines referencing non-existent host keys.


justin-stephenson (approved):
Ack, system tests running locally after building sssd-ci-containers with this PR.

PR #107, opened 2024-09-27
provision: Add binutils to have strings available on client

jakub-vavra-cz:
Strings command is used in feature detection on client but binutils providing it are by default missing in containers.


sumit-bose (approved):
Hi,

thank you for adding `binutils`, ACK.

bye,
Sumit

PR #111, opened 2024-11-21
Retry IPA client install

justin-stephenson:
ipa-client-install with TAG=fedora-42 can fail in PRCI with
"Unable to find IPA Server to join" although SRV records are
present and valid.

```
- 2024-11-20T21:15:38Z DEBUG [IPA Discovery]
- 2024-11-20T21:15:38Z DEBUG Starting IPA discovery with domain=ipa.test, servers=None, hostname=client.test
- 2024-11-20T21:15:38Z DEBUG Search for LDAP SRV record in ipa.test
- 2024-11-20T21:15:38Z DEBUG Search DNS for SRV record of _ldap._tcp.ipa.test
- 2024-11-20T21:15:38Z DEBUG DNS record not found: NXDOMAIN'
- 2024-11-20T21:15:38Z DEBUG No LDAP server found
```

### DIFF
--- a/src/ansible/roles/client/tasks/enroll_IPA.yml
+++ b/src/ansible/roles/client/tasks/enroll_IPA.yml
@@ -11,6 +11,10 @@
       --domain {{ ipa_domain | quote }}                \
       --principal admin                                \
       --password {{ ipa_password | quote }}
+  # Retry to workaround "Unable to find IPA Server to join" failure (fedora-42)
+  # in PRCI fedora build workflow
+  retries: 2
+  delay: 10
   args:
     creates: /etc/ipa/ca.crt
 


thalman (approved):
LGTM, ACK

ikerexxe:
@justin-stephenson is the race condition that you are trying to workaround reported anywhere (i.e. pagure, github, rhbz)?

justin-stephenson:
> @justin-stephenson is the race condition that you are trying to workaround reported anywhere (i.e. pagure, github, rhbz)?

No, I merely suspect this is a race condition based on the symptoms of the issue and the resolution. It also happens only in PRCI as I could not reproduce locally. I don't really even know which component is at fault here (dnsmasq, container tools networking?)

ikerexxe (approved):
> No, I merely suspect this is a race condition based on the symptoms of the issue and the resolution. It also happens only in PRCI as I could not reproduce locally. I don't really even know which component is at fault here (dnsmasq, container tools networking?)

mmmm I see. My fear is that this workaround will mask the error and it'll never get fixed. If by any chance you have the logs I would appreciate it if you could open a ticket on FreeIPA, so that we can at least have an initial discussion about what the problem is.

LGTM!

justin-stephenson:
> > No, I merely suspect this is a race condition based on the symptoms of the issue and the resolution. It also happens only in PRCI as I could not reproduce locally. I don't really even know which component is at fault here (dnsmasq, container tools networking?)
> 
> mmmm I see. My fear is that this workaround will mask the error and it'll never get fixed. If by any chance you have the logs I would appreciate it if you could open a ticket on FreeIPA, so that we can at least have an initial discussion about what the problem is.
>
At this moment I don't know if this issue is caused by FreeIPA, or the DNS server on `ci-dns` container, or container networking state. I'll troubleshoot this further to try to understand the root cause, but at a lower priority.

PR #115, opened 2025-01-30
Update for multiarch testing

jakub-vavra-cz:
We need to disable beaker tasks repo for sssd package installation to prevent stray package inclusion.
Rsync is missing from RHEL 10.0 aarch64 and needs to be explicitly installed for common role.


pbrezina:
Ack.

pbrezina (approved):

PR #121, opened 2025-03-22
ipasmartcard: new role for IPA smart card support

spoore1:
Configure IPA server and client using ipa-advise scripts.

Currently there is an issue with IPA WebUI OCSP verifications when the IPA server is configured using an external CA.  To work around this we disable OCSP checking in /etc/httpd/conf.d/ssl.conf on the IPA server.

### DIFF
--- a/src/ansible/playbook_image_service.yml
+++ b/src/ansible/playbook_image_service.yml
@@ -21,6 +21,7 @@
   roles:
   - ipa
   - { role: passkey, when: ansible_distribution == "Ubuntu" or ansible_distribution == "Debian" }
+  - { role: ipasmartcard, when: ansible_os_family == 'RedHat' and virt_smartcard }
 
 - hosts: client
   gather_facts: no
@@ -28,6 +29,7 @@
   - client
   - { role: passkey, when: ansible_distribution == "Ubuntu" or ansible_distribution == "Debian" }
   - { role: virtsmartcard, when: ansible_distribution != "Ubuntu" and ansible_distribution != "Debian" and virt_smartcard }
+  - { role: ipasmartcard, when: ansible_os_family == 'RedHat' and virt_smartcard }
 
 - hosts: nfs
   gather_facts: no
--- a/src/ansible/roles/ipasmartcard/tasks/main.yml
+++ b/src/ansible/roles/ipasmartcard/tasks/main.yml
@@ -1,0 +1,75 @@
+- name: Checking DNS resolution first
+  shell: |
+    dig ipa-ca.{{ ipa_domain }} | grep NOERROR
+  register: result
+  until: result.rc == 0
+  retries: 10
+  delay: 2
+
+- name: Make sure virtual smartcard dir exists
+  file:
+    path: "{{ virt_smartcard_dir }}"
+    state: directory
+
+- block:
+  - name: Run IPA advise script for server setup for smart cards
+    shell: |
+      kinit admin@{{ ipa_domain.upper() }}
+      ipa-advise config-server-for-smart-card-auth > {{ virt_smartcard_dir }}/sc_server.sh
+      sh -x {{ virt_smartcard_dir }}/sc_server.sh {{ sssd_ca_path }} {{ ipa_ca_path }}
+    args:
+      stdin: '{{ ipa_password }}'
+    register: ipa_advise_sc_server
+
+  - name: Workaround to disable IPA WebUI OCSP checking
+    lineinfile:
+      path: /etc/httpd/conf.d/ssl.conf
+      regexp: 'SSLOCSPEnable on\s*$'
+      line: 'SSLOCSPEnable on no_ocsp_for_cert_ok'
+
+  - name: Restart httpd service
+    ansible.builtin.systemd_service:
+      name: httpd
+      state: restarted
+  when: inventory_hostname == groups.ipa.0
+
+- block:
+  - name: Get IPA advise script for client setup for smart cards from IPA server
+    command: ipa-advise config-client-for-smart-card-auth
+    delegate_to: "{{ groups.ipa.0 }}"
+    register: ipa_advise_sc_client_script
+
+  - name: Write advise script to sc_client.sh
+    copy:
+      content: "{{ ipa_advise_sc_client_script.stdout }}"
+      dest: "{{ virt_smartcard_dir }}/sc_client.sh"
+
+  - name: Get IPA CA certificate from IPA server
+    command: cat /root/ipa.crt
+    delegate_to: "{{ groups.ipa.0 }}"
+    register: ipa_ca_cert
+
+  - name: Write IPA CA certificate to /root/ipa.crt
+    copy:
+      content: "{{ ipa_ca_cert.stdout }}"
+      dest: /root/ipa.crt
+
+  - name: Add krb5.keytab link for IPA client
+    file:
+      src: /var/enrollment/{{ ipa_domain }}.keytab
+      dest: /etc/krb5.keytab
+      state: link
+
+  - name: Run IPA advise script on client
+    shell: |
+      kinit admin@{{ ipa_domain.upper() }}
+      sh -x {{ virt_smartcard_dir }}/sc_client.sh {{ sssd_ca_path }} {{ ipa_ca_path }}
+    args:
+      stdin: "{{ ipa_password }}"
+    register: ipa_advise_sc_client
+
+  - name: Remove krb5.keytab link from IPA client
+    file:
+      path: /etc/krb5.keytab
+      state: absent
+  when: inventory_hostname == groups.client.0
--- a/src/ansible/roles/ipasmartcard/defaults/main.yml
+++ b/src/ansible/roles/ipasmartcard/defaults/main.yml
@@ -1,0 +1,4 @@
+ipa_ca_path: /root/ipa.crt
+sssd_ca_path: /etc/pki/ca-trust/source/anchors/ca.crt
+ipa_password: "{{ hostvars[groups.ipa.0].ansible_password | default(service.ipa.password) }}"
+ipa_domain: "{{ hostvars[groups.ipa.0].ipa_domain | default(service.ipa.domain) }}"


src/ansible/roles/ipasmartcard/tasks/main.yml
```diff
@@ -0,0 +1,68 @@
...
+      regexp: 'SSLOCSPEnable on'
+      line: 'SSLOCSPEnable on no_ocsp_for_cert_ok'
+      backrefs: yes
+
+  - name: Restart httpd service
+    ansible.builtin.systemd_service:
+      name: httpd
+      state: restarted
+
+  when: inventory_hostname == 'master.ipa.test'
```
jakub-vavra-cz, line 24: I would target group ipa instead and make sure that it uses variables where appropriate (like kinit).

src/ansible/roles/ipasmartcard/tasks/main.yml
```diff
@@ -0,0 +1,68 @@
...
+
+- block:
+  ################# IPA Client ####################
+  - name: Get IPA advise script for client setup for smart cards from IPA server
+    shell: |
+      ipa-advise config-client-for-smart-card-auth
+    delegate_to: "master.ipa.test"
```
jakub-vavra-cz, line 31: Another hardcoded  host.  If we want to setup client with first ipa it would be better to reference it though group ipa nad index instead.

src/ansible/roles/ipasmartcard/tasks/main.yml
```diff
@@ -0,0 +1,68 @@
...
+    args:
+      stdin: '{{ ipa_password }}'
+    register: ipa_advise_sc_client
+
+  - name: Remove krb5.keytab link from IPA client
+    file:
+      path: /etc/krb5.keytab
+      state: absent
+
+  when: inventory_hostname == 'client.test'
```
jakub-vavra-cz, line 68: This should point to client group/role.

jakub-vavra-cz:
I would rather see groups used instead of hardcoded hostnames.

src/ansible/roles/ipasmartcard/tasks/main.yml
spoore1: That was one of the things I wanted input on.   Take a look at the update and let me know if I missed anything or if there's a better (or more standard to the project) way to handle this.

src/ansible/roles/ipasmartcard/tasks/main.yml
```diff
@@ -0,0 +1,68 @@
...
+    register: ipa_ca_cert
+
+  - name: Write IPA CA certificate to /root/ipa.crt
+    copy:
+      content: "{{ ipa_ca_cert.stdout }}"
+      dest: /root/ipa.crt
+
+  - name: Add krb5.keytab link for IPA client
+    file:
+      src: /enrollment/ipa.test.keytab
```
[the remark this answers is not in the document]
spoore1: Any idea when that may be reviewed and merged?    Maybe I should switch this now and add a separate temp commit that allows the checks to pass by creating a link /enrollment -> /data/enrollment?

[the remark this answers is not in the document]
spoore1: Oh, it looks like there's already a link created in the client role:
https://github.com/SSSD/sssd-ci-containers/pull/117/files#diff-506a7085e88a3bf0585d0606e4dcbf709b126621c449269564eabd8401787650R19

src/ansible/playbook_image_service.yml
```diff
@@ -21,13 +21,15 @@
...
   - { role: passkey, when: ansible_distribution == "Ubuntu" or ansible_distribution == "Debian" }
+  - { role: ipasmartcard, when: ansible_distribution != "Ubuntu" and ansible_distribution != "Debian" and virt_smartcard }
 
 - hosts: client
   gather_facts: no
   roles:
   - client
   - { role: passkey, when: ansible_distribution == "Ubuntu" or ansible_distribution == "Debian" }
   - { role: virtsmartcard, when: ansible_distribution != "Ubuntu" and ansible_distribution != "Debian" and virt_smartcard }
+  - { role: ipasmartcard, when: ansible_distribution != "Ubuntu" and ansible_distribution != "Debian" and virt_smartcard }
```
thalman, line 32: the same as above

src/ansible/roles/ipasmartcard/tasks/main.yml
```diff
@@ -0,0 +1,72 @@
...
+
+  - name: Restart httpd service
+    ansible.builtin.systemd_service:
+      name: httpd
+      state: restarted
+  when: inventory_hostname == groups.ipa.0
+
+- block:
+  - name: Get IPA advise script for client setup for smart cards from IPA server
+    shell: |
```
jakub-vavra-cz, line 34: Would command: work here?

src/ansible/roles/ipasmartcard/tasks/main.yml
```diff
@@ -0,0 +1,72 @@
...
+  until: result.rc == 0
+  retries: 10
+  delay: 10
+
+- block:
+  - name: Run IPA advise script for server setup for smart cards
+    shell: |
+      kinit admin@{{ ipa_domain.upper() }}
+      ipa-advise config-server-for-smart-card-auth > /tmp/sc_server.sh
+      sh -x /tmp/sc_server.sh {{ sssd_ca_path }} {{ ipa_ca_path }}
```
jakub-vavra-cz, line 14: It might be better to place this script elsewhere. Stig mounts /tmp as noexec not sure about fips.

src/ansible/roles/ipasmartcard/tasks/main.yml
```diff
@@ -0,0 +1,72 @@
...
+    register: ipa_ca_cert
+
+  - name: Write IPA CA certificate to /root/ipa.crt
+    copy:
+      content: "{{ ipa_ca_cert.stdout }}"
+      dest: /root/ipa.crt
+
+  - name: Add krb5.keytab link for IPA client
+    file:
+      src: /enrollment/{{ ipa_domain }}.keytab
```
jakub-vavra-cz, line 56: The change to /var/enrollment is already merged.

jakub-vavra-cz (changes requested):
See inline

src/ansible/playbook_image_service.yml
```diff
@@ -21,13 +21,15 @@
   roles:
   - ipa
   - { role: passkey, when: ansible_distribution == "Ubuntu" or ansible_distribution == "Debian" }
+  - { role: ipasmartcard, when: ansible_distribution != "Ubuntu" and ansible_distribution != "Debian" and virt_smartcard }
```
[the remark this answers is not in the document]
spoore1: What about using `ansible_facts[‘os_family’] == 'RedHat' `?

src/ansible/playbook_image_service.yml
spoore1: Same as above.  Is os_family acceptable there?

src/ansible/roles/ipasmartcard/tasks/main.yml
spoore1: Yes, I think so.   Will try with that to be sure.

src/ansible/roles/ipasmartcard/tasks/main.yml
spoore1: How about I use the virt_smartcard_dir variable?  /opt/test_ca should be good right?

jakub-vavra-cz: Yes. That could work. Alternatively I would also consider moving everything under /var if we went to run smartcard tests in image mode in the future.

src/ansible/playbook_image_service.yml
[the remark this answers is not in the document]
spoore1: switched to ansible_os_family.   The ansible_facts variable wasn't available because gather_facts is no.

src/ansible/roles/ipasmartcard/tasks/main.yml
spoore1: Let's stick with this for now if that's ok.   If we need to run smartcard tests in image mode, I'll submit a PR to update all smart card related paths.

src/ansible/roles/ipasmartcard/tasks/main.yml
```diff
@@ -0,0 +1,76 @@
...
+      sh -x {{ virt_smartcard_dir }}/sc_server.sh {{ sssd_ca_path }} {{ ipa_ca_path }}
+    args:
+      stdin: '{{ ipa_password }}'
+    register: ipa_advise_sc_server
+
+  - name: Workaround to disable IPA WebUI OCSP checking
+    lineinfile:
+      path: /etc/httpd/conf.d/ssl.conf
+      regexp: 'SSLOCSPEnable on'
+      line: 'SSLOCSPEnable on no_ocsp_for_cert_ok'
```
thalman, line 28: Regexp match the line so if you run this two times it might not work correctly.

`regexp: SSLOCSPEnable on\\s*$` 

might be better.

src/ansible/roles/ipasmartcard/tasks/main.yml
```diff
@@ -0,0 +1,76 @@
...
+      backrefs: yes
```
thalman, line 29: backrefs are not needed here

src/ansible/roles/ipasmartcard/tasks/main.yml
```diff
@@ -0,0 +1,76 @@
+- name: Checking DNS resolution first
+  shell: |
+    dig ipa-ca.{{ ipa_domain }} | grep NOERROR
+  register: result
+  until: result.rc == 0
+  retries: 10
+  delay: 10
```
thalman, line 7: I wonder if the delay can be shorter (1 ~ 2s). How much time does this step usually take to settle down?

thalman (changes requested):

src/ansible/roles/ipasmartcard/tasks/main.yml
spoore1: I'd have to dig back through some old failures but, I thought it was almost 30 seconds but, considering that there are other steps between when the IPA server is setup and this, that probably pads a little extra time.   So, I'll switch this to 2 seconds and we can adjust to higher only if needed?

src/ansible/roles/ipasmartcard/tasks/main.yml
spoore1: Ah good point.   I'll adjust the regexp.

src/ansible/roles/ipasmartcard/tasks/main.yml
spoore1: dropping

jakub-vavra-cz (approved):

thalman (approved):
Thanks for the update

PR #125, opened 2025-04-14
 adding second network to ipa2 domain

danlavu:


danlavu:
I tried sourcing tools/get-container-engine.sh, it seems setting it to ./src/tools works locally but fails in ci. Then ./tools/ works in CI but fails locally, or it maybe the other way around. I updated it so it doesn't source the file.

pbrezina:
Is this deprecated by https://github.com/SSSD/sssd-ci-containers/pull/129?

PR #126, opened 2025-04-22
Skip ipasmartcard on client if not joined to IPA

spoore1:
When configured using playbook_vm.yaml, the client is not enrolled in IPA by default.  This causes the ipasmartcard role to currently fail to find the IPA kerberos keytab needed.   The client side setup can be skipped in the case where the system is not enrolled to IPA.

### DIFF
--- a/src/ansible/roles/dns/templates/etc.dnsmasq.conf.j2
+++ b/src/ansible/roles/dns/templates/etc.dnsmasq.conf.j2
@@ -14,7 +14,7 @@ cache-size=0
 
 # These zones have their own DNS server
 {% for host in groups['ipa'] %}
-server=/{{ hostvars[host]['ansible_facts']['domain'] }}/{{ hostvars[host]['ansible_facts']['default_ipv4']['address'] }}
+server=/{{ '.'.join(hostvars[host].inventory_hostname.split('.')[1:]) }}/{{ hostvars[host]['ansible_facts']['default_ipv4']['address'] }}
 {% endfor %}
 {% if 'dc.samba.test' in hostvars %}
 server=/samba.test/{{ hostvars['dc.samba.test']['ansible_facts']['default_ipv4']['address'] }}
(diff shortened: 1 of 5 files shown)


thalman (approved):
LGTM

madhuriupadhye:
Hello @spoore1,

I see DNS issue while checking in idm-ci,
```
2025-04-23T06:44:46 TASK [ipasmartcard : Checking DNS resolution first] ****************************
2025-04-23T06:45:22 FAILED - RETRYING: [client.test]: Checking DNS resolution first (10 retries left).
```

src/ansible/roles/dns/templates/etc.dnsmasq.conf.j2
```diff
@@ -13,9 +13,12 @@ domain=test
 cache-size=0
 
 # These zones have their own DNS server
-{% for host in groups['ipa'] %}
-server=/{{ hostvars[host]['ansible_facts']['domain'] }}/{{ hostvars[host]['ansible_facts']['default_ipv4']['address'] }}
-{% endfor %}
+{% if 'master.ipa.test' in hostvars %}
+server=/ipa.test/{{ hostvars['master.ipa.test']['ansible_facts']['default_ipv4']['address'] }}
```
jakub-vavra-cz, line 17: Why You did not use the meta_domain variable I suggested?

spoore1: I did originally but, Justin brought up a good point about that variable being available outside of our ci.  I don't think we use these outside of that but currently, but, it was easier to just use the same format as the rest of the file which relies already on some hardcoded entries.

madhuriupadhye:
Tested and worked for me in IDM-CI.

madhuriupadhye (approved):
LGTM

src/ansible/roles/dns/templates/etc.dnsmasq.conf.j2
jakub-vavra-cz: Alternatively there is this code that might be used as base:
https://github.com/SSSD/sssd-ci-containers/blob/master/src/ansible/roles/ipa/tasks/main.yml#L13

PR #127, opened 2025-04-23
DNS: Removed an extra 'if'

aplopez:
I failed to test it on my environment because of several other errors produced by missing tools. PR tests look good.


jakub-vavra-cz (approved):
LGTM

PR #129, opened 2025-05-02
 moving ipa2 to a second subnet 

danlavu:
* making this environment more realistic
* moved the network create to use the docker command because docker-compose defaults the isolation value to true

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -69,8 +69,8 @@ services:
     - label=disable
     - seccomp=unconfined
     networks:
-      sssd:
-        ipv4_address: 172.16.100.11
+      ipa:
+        ipv4_address: 172.16.110.10
   ldap:
     image: ${REGISTRY}/ci-ldap:${TAG}
     container_name: ldap
@@ -212,10 +212,7 @@ services:
 networks:
   sssd:
     name: sssd-ci
-    driver: bridge
-    ipam:
-     config:
-       - subnet: 172.16.100.0/24
-         gateway: 172.16.100.1
-     options:
-        driver: host-local
+    external: true
+  ipa:
+    name: ipa-ci
+    external: true
--- a/src/build.sh
+++ b/src/build.sh
@@ -50,7 +50,15 @@ function cleanup {
 }
 
 function compose {
-  docker-compose -f "../docker-compose.yml" -f "./docker-compose.build.yml" $@
+  if [ $@ == "up" ]; then
+    ./tools/create-networks.sh
+    docker-compose -f "../docker-compose.yml" -f "./docker-compose.build.yml" $@ --detach
+  elif [ "$@" == "down" ]; then
+    docker-compose -f "../docker-compose.yml" -f "./docker-compose.build.yml" $@
+    ./tools/remove-networks.sh
+  else
+    docker-compose -f "../docker-compose.yml" -f "./docker-compose.build.yml" $@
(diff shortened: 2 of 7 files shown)


danlavu:
This should work, @justin-stephenson, can you please test this? Something may be foo bar on my workstation, but I think it should work. The second network is created and master.ipa2.test is provisioned correctly. The problem is that 'sssd' network is not routable to the 'ipa' network.

However, both the 'sssd' and 'IPA' networks are routable from the host and the ad vagrant network.

justin-stephenson:
> This should work, @justin-stephenson, can you please test this? Something may be foo bar on my workstation, but I think it should work. The second network is created and master.ipa2.test is provisioned correctly. The problem is that 'sssd' network is not routable to the 'ipa' network.
> 
> However, both the 'sssd' and 'IPA' networks are routable from the host and the ad vagrant network.

Hi,

I built this PR but cannot ping `master.ipa2.test` or `172.16.110.10` from `master.ipa.test` , I see some errors about dns in the journal.

```
May 05 15:26:10 fedora.test.local dns[856731]: dnsmasq[60]: query[A] master.ipa2.test from 172.16.100.1
May 05 15:26:15 fedora.test.local aardvark-dns[856608]: 5346 dns request got empty response

```

Sorry I can't help much here I don't have much experience with networking

danlavu:
Thanks for testing, it's not my workstation, @jakub-vavra-cz suggested something that I will look into.

docker-compose.yml
```diff
@@ -198,3 +198,16 @@ networks:
...
+          com.docker.network.bridge.enable_icc=true
+  ipa:
+    name: ipa-ci
+    driver: bridge
+    ipam:
+     config:
+       - subnet: 172.16.110.0/24
+         gateway: 172.16.110.1
+     options:
+        driver: host-local
```
thalman, line 211: What this diver option does? Looks like good candidate for blocking the trafic.

thalman: Ah OK this should be fine

danlavu:
After talking to @thalman and seeing the ci failures. ```podman-compose``` isn't installed on other distros. I'm going to create the networks using the docker command and have docker-compose use pre-existing networks, so there is no dependency on podman-compose.

pbrezina:
Bump, any progress here?

src/build.sh
```diff
@@ -47,10 +47,12 @@ set -xe
 function cleanup {
   ${DOCKER} rm sssd-wip-base --force || :
   compose down
+  ./tools/remove-networks.sh
 }
 
 function compose {
-  docker-compose -f "../docker-compose.yml" -f "./docker-compose.build.yml" $@
+  ./tools/create-networks.sh
+  docker-compose -f "./docker-compose.yml" -f "./docker-compose.keycloak.yml" -f "./docker-compose.build.yml" $@
```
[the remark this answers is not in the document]
danlavu: Fixed.

danlavu:
Okay, being unfamiliar with the Makefile there were a number of mistakes. I figured them out, there are enough changes from Tomas's review that it should be reviewed again. The containers are now being built successfully.

* added make start (doesn't create the networks), in the event you just wanted to start the containers from being stopped.
* The compose function has a condition for up|down to execute the script

justin-stephenson:
By the way @pbrezina may want to look and review this PR.

PR #138, opened 2025-07-07
Enable virt_smartcard for centos10

spoore1:
virt_smartcard was explicitly disabled for Centos 10 setup.   This needs
to be enabled so that the centos10 system tests can run smart card tests
in PR-CI.   Removing this from the Centos10 specific facts task file.

Also fix a virt_smartcard typo in playbook_vm.yml


jakub-vavra-cz (approved):
LGTM

PR #141, opened 2025-08-12
pqc: add pqc certificates

pbrezina:


pbrezina:
Currently, LDAP starts with PQC certificates but it disables SSL since it can not used them. Dropping this PR for now.

PR #142, opened 2025-08-18
install chocolately and bind tool package

danlavu:


jakub-vavra-cz (approved):
Looks good, but please make sure that it works also on Win 2012R2 before merging.

danlavu:
> SEE comment

ACK, I split the play into two. I moved installing chocolately to a task before another reboot.

PR #146, opened 2025-09-10
Revert "install chocolately and bind tool package"

danlavu:
Reverts SSSD/sssd-ci-containers#142


jakub-vavra-cz (approved):
LGTM. I would no mind if some retries and ignore_error were included instead through.

danlavu:
We no longer need bind on AD, so this patch can be dropped entirely.

PR #147, opened 2025-09-16
PRCI fixes

justin-stephenson:


src/ansible/roles/packages/tasks/Fedora.yml
```diff
@@ -233,11 +233,32 @@
 
 - name: Install packages for IPA base image
   block:
+  # With additional repos enabled on c8s, installing IPA fails with
+  # Error: No available modular metadata for modular package
+  # 'jaxb-fi-2.1.0-1.module_el8+321+a8d35a21.noarch'
+  - name: Disable crb/buildroot repos due to CentOS stream 8 error
```
alexey-tikhonov, line 239: Do we still build C8S container? Is it ever possible and what for?

justin-stephenson: It was last built successfully 2 months ago according to https://quay.io/repository/sssd/ci-base-ldap?tab=tags - I don't know if it is being used at all.

alexey-tikhonov: IIUC, c8s repos aren't up-to-date even if available.
We don't use it even in sssd-2-9-4 branch: https://github.com/SSSD/sssd/blob/sssd-2-9-4/contrib/ci/get-matrix.py
I would better stop building c8s container and removed all related code.

justin-stephenson: Okay would you prefer to remove 'c8s' and debian as part of this PR also? CC @pbrezina

justin-stephenson: I added a new commit to drop c8s, let's wait for CI results.

justin-stephenson: centos8 is removed now with CI passing.

alexey-tikhonov (approved):

pbrezina (approved):

PR #150, opened 2025-10-07
sss pac responder client

sumit-bose:


pbrezina:
> > Maybe we can just remove the sssd-ci-random package and rename it to sssd-ci[-test-deps]?
> 
> Hi,
> 
> do I understand correctly that it would be better to rename the rpm in a separate PR and wait until it is committed and a new package is build in copr before it can be used here?

Yes, if we want to make CI green. But the RPM is tested in "Build RPMS" job, so I do not think it is necessary.

pbrezina (approved):

PR #163, opened 2026-02-10
Changes for dyndns tests.

danlavu:
- The genericDNS method requires `dig` to be installed and it's missing upstream but not downstream.
- Adding a static reverse zone for each topology. 
- Fixing IPA zones that are not dyndns enabled
- Changing the hostname on realm and ipa client install, the keytabs will now reflect the topology fqn instead of client.test


data/configs/dnsmasq.conf
```diff
@@ -17,7 +17,9 @@ server=/samba.test/172.16.100.30
 server=/ad.test/172.16.200.10
 
 # Add reverse zones for artificial hosts in IPA domain
-server=/251.255.10.in-addr.arpa/172.16.100.10
+rev-server=10.255.250.0/24,172.16.200.10 # ad
+rev-server=10.255.251.0/24,172.16.100.10 # ipa
+rev-server=10.255.252.0/24,172.16.100.30 # samba
```
justin-stephenson: Can you clarify how are these "artificial hosts" used and why do we need reverse zones for AD/Samba?

src/ansible/roles/client/tasks/enroll_samba.yml
```diff
@@ -19,7 +19,9 @@
     mode: 0644
 
 - name: Realm join Samba domain
-  command: realm join {{ samba_domain | quote }} --verbose
+  shell: |
+    hostname {{ ansible_hostname + '.' + samba_domain | quote}}
+    realm join {{ samba_domain | quote }} --verbose
```
jakub-vavra-cz: I am not a fan of putting shell everywhere just to inline crap in one go.
These should be two separate ansible commands.

src/ansible/roles/dns/templates/etc.dnsmasq.conf.j2
```diff
@@ -27,15 +27,47 @@ server=/{{ hostvars[ad]['ansible_facts']['windows_domain'] }}/{{ hostvars[ad]['a
...
+# Add reverse zones for artificial hosts
 {% if 'master.ipa.test' in hostvars %}
-# Add reverse zones for artificial hosts in IPA domain
-server=/251.255.10.in-addr.arpa/{{ hostvars['master.ipa.test']['ansible_facts']['default_ipv4']['address'] }}
+rev-server=10.255.251.0/24,{{ hostvars['master.ipa.test']['ansible_facts']['default_ipv4']['address'] }}
+{% endif %}
+{% if 'dc.samba.test' in hostvars %}
+rev-server=10.255.252.0/24,{{ hostvars['dc.samba.test']['ansible_facts']['default_ipv4']['address'] }}
+{% endif %}
+{% if 'dc.ad.test' in hostvars %}
```
jakub-vavra-cz: There should not be any dc.ad.test in 90% cases as (apart from passkey) everywhere should be the domain randomized.

src/ansible/roles/client/tasks/enroll_AD.yml
```diff
@@ -18,6 +18,10 @@
     group: root
     mode: 0644
 
+- name: Set hostname to FQDN for AD join
+  ansible.builtin.command:
+    name: "hostname {{ inventory_hostname }}.{{ ad_domain }}"
```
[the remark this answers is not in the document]
danlavu:
```

## Available containers

| Name         |        IP       | FQDN                   | Description                         |
|--------------|-----------------|------------------------|-------------------------------------|
| ipa          | `172.16.100.10` | `master.ipa.test`      | IPA server                          |
| ldap         | `172.16.100.20` | `master.ldap.test`     | TLS ready 389 Directory Server      |
| samba        | `172.16.100.30` | `dc.samba.test`        | Samba DC root domain                |
| client       | `172.16.100.40` | `client.$domain.test`  | Client machine with configured SSSD |
| nfs          | `172.16.100.50` | `nfs.test`             | NFS server                          |
| kdc          | `172.16.100.60` | `kdc.test`             | Kerberos KDC                        |
| keycloak     | `172.16.100.70` | `master.keycloak.test` | Keycloak IdP                        |
| ipa2         | `172.16.100.11` | `master.ipa2.test`     | IPA server in different realm       |

* Note, the client machine hostname is `client.$domain.test`, where `$domain` is the domain of the
  connected provider, e.g. `ldap`, `ipa` or `samba`. It is always accessible by `client.test`
  hostname as well.
  ```
  
   ?

justin-stephenson (approved):
Ack, thank you.

jakub-vavra-cz (changes requested):
Just one inline comment, otherwise LGTM.

jakub-vavra-cz (changes requested):
My comment about src/ansible/roles/client/tasks/enroll_samba.yml is not resolved.

jakub-vavra-cz (approved):
LGTM

thalman (approved):
Looks good, thanks

justin-stephenson (approved):
Ack from me, it would be great to test this by pushing images to an alternate registry at quay.io

thalman:
I just went trough it once again to check last pushes. Looks good, ACK

PR #178, opened 2026-05-19
rpms: Use git describe for versioning ci-deps

justin-stephenson:
Workaround for https://github.com/fedora-copr/copr/issues/4065 which is caused by recent versions of mock performing multiple srpm builds.

Fix for failling ci-sssd builds of https://copr.fedorainfracloud.org/coprs/g/sssd/ci-deps/package/ci-sssd/

### DIFF
--- a/src/rpms/Makefile
+++ b/src/rpms/Makefile
@@ -1,3 +1,13 @@
+set-version: prep
+	@dnf install -y git
+	@if git describe; then \
+		GIT_VERSION=$$(git describe | sed 's/-/_/g'); \
+		echo "Git describe succeeded. Version: $$GIT_VERSION"; \
+		sed -i "s/^Release:.*/Release: $$GIT_VERSION%{?dist}/" $(PWD)/rpmbuild/SPECS/ci-sssd.spec; \
+	else \
+		echo "Git describe failed with exit code $$?"; \
+	fi
+
 prep:
 	mkdir -p $(PWD)/rpmbuild/BUILD
 	mkdir -p $(PWD)/rpmbuild/RPMS
@@ -8,11 +18,11 @@ prep:
 	cp sss_pac_responder_client.c $(PWD)/rpmbuild/SOURCES
 	cp ci-sssd.spec $(PWD)/rpmbuild/SPECS
 
-srpm: prep
-	rpmbuild --define "_topdir $(PWD)/rpmbuild" -bs ci-sssd.spec
+srpm: set-version
+	rpmbuild --define "_topdir $(PWD)/rpmbuild" -bs $(PWD)/rpmbuild/SPECS/ci-sssd.spec
 
-rpms: prep
-	rpmbuild --define "_topdir $(PWD)/rpmbuild" -ba ci-sssd.spec
+rpms: set-version
+	rpmbuild --define "_topdir $(PWD)/rpmbuild" -ba $(PWD)/rpmbuild/SPECS/ci-sssd.spec
 
 clean:
 	rm -fr $(PWD)/rpmbuild
--- a/src/rpms/ci-sssd.spec
+++ b/src/rpms/ci-sssd.spec
@@ -1,8 +1,6 @@
-%{!?build_timestamp:%global build_timestamp %(date +"%%Y_%%m_%%d_%%H_%%M_%%S")}
-
 Name:           ci-sssd
 Version:        1
-Release:        2%{?dist}.%{build_timestamp}
+Release:        2%{?dist}
 Summary:        SSSD CI Packages
 URL:            https://github.com/SSSD/sssd-ci-containers
 


src/rpms/ci-sssd.spec
```diff
@@ -1,8 +1,8 @@
-%{!?build_timestamp:%global build_timestamp %(date +"%%Y_%%m_%%d_%%H_%%M_%%S")}
+%global git_commit_ts %(git log -1 --date=format:"%%Y_%%m_%%d_%%H_%%M_%%S" --format="%%ad" || date +"%%Y_%%m_%%d_%%H_%%M_%%S")
```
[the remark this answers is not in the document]
justin-stephenson: Fixed as suggested.

pbrezina:
How about using `git describe`? We would need to add at least one tag to the repository, but that is not an issue. I added local tag 1.0.0 to HEAD~8 (not annotated thus required to use --tags). It then prints tag, number of commits since, last commit sha.

```
$ git describe --tags
1.0.0-8-gf0ef010
```

pbrezina:
Maybe even use %autorelease and %autochangelog

justin-stephenson:
> Maybe even use %autorelease and %autochangelog

Thanks, I went with this approach.

Builds are now successful tested with https://copr.fedorainfracloud.org/coprs/jstephen/sssd-ci-deps-testing/builds/

And testing the updated rpms are observed by dnf as 'latest'

```
# dnf copr enable @sssd/ci-deps 
# dnf copr enable jstephen/sssd-ci-deps-testing 

[root@vm-10-0-184-190 ~]# dnf install ci-sssd-test-deps
Updating and loading repositories:
Repositories loaded.
Package                                                                                    Arch          Version                                                                                     Repository                                                                        Size
Installing:
 ci-sssd-test-deps                                                                         x86_64        0:1-3.fc44                                                                                  copr:copr.fedorainfracloud.org:jstephen:sssd-ci-deps-testing                  25.5 KiB

Transaction Summary:
 Installing:         1 package

Total size of inbound packages is 13 KiB. Need to download 13 KiB.
After this operation, 25 KiB extra will be used (install 25 KiB, remove 0 B).
Is this ok [y/N]: 
```

justin-stephenson:
I still need to test that new COPR builds (generated from new commit pushed to sssd-ci-containers) will increment the NVR with %autorelease, please standby

justin-stephenson:
> I still need to test that new COPR builds (generated from new commit pushed to sssd-ci-containers) will increment the NVR with %autorelease, please standby

It did not increase `release` automatically (it seems that %autorelease commit counting [does not work with COPR builds](https://docs.pagure.org/copr.copr/user_documentation.html#faq-autospec)), so I added the date suffix (only YY_MM_DD). `ci-sssd-1-3.2026_05_20.fc44.src.rpm`. This works but doesn't really take advantage of %autorelease, maybe `git describe` is the way to go here.

justin-stephenson:
I switched this PR to use `git describe` in the Makefile but for the `Build RPMs` Github action to succeed it also needs to check if inside of a git repository, I added support for this also.

Tag will need to be created either with a greater NVR than `1-2.fc44.2025_12_09_14_01_08` for COPR/dnf to detect the new rpm as the latest. (Or remove all previous `ci-sssd-test-deps` build from COPR?)

```
justin@justin-fedora:~/github/sssd$ rpmdev-vercmp 1-3.0.0_2_g2fc2e57 1-2.2025_12_09_14_00_20
1-3.0.0_2_g2fc2e57 > 1-2.2025_12_09_14_00_20

```

Tested succesfully here:

https://copr.fedorainfracloud.org/coprs/jstephen/sssd-ci-deps-testing/builds/

src/rpms/Makefile
```diff
@@ -1,3 +1,13 @@
+set-version:
+	@dnf install -y git
+	@if git describe; then \
+		GIT_VERSION=$$(git describe | sed 's/-/_/g'); \
+		echo "Git describe succeeded. Version: $$GIT_VERSION"; \
+		sed -i "s/^Release:.*/Release: $$GIT_VERSION/" ci-sssd.spec; \
+	else \
+		echo "Git describe failed with exit code $$?"; \
+	fi
```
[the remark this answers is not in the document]
justin-stephenson: Fixed the `2` and `3` issues here but `dnf install -y git` is necessary or COPR build will fail.

src/rpms/Makefile
```diff
@@ -8,10 +18,10 @@ prep:
 	cp sss_pac_responder_client.c $(PWD)/rpmbuild/SOURCES
 	cp ci-sssd.spec $(PWD)/rpmbuild/SPECS
 
-srpm: prep
+srpm: prep set-version
 	rpmbuild --define "_topdir $(PWD)/rpmbuild" -bs ci-sssd.spec
 
-rpms: prep
+rpms: prep set-version
 	rpmbuild --define "_topdir $(PWD)/rpmbuild" -ba ci-sssd.spec
```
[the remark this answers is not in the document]
justin-stephenson: Fixed.

src/rpms/Makefile
```diff
@@ -1,3 +1,13 @@
+set-version: prep
+	@dnf install -y git
+	@if git describe; then \
+		GIT_VERSION=$$(git describe | sed 's/-/_/g'); \
+		echo "Git describe succeeded. Version: $$GIT_VERSION"; \
+		sed -i "s/^Release:.*/Release: $$GIT_VERSION/" $(PWD)/rpmbuild/SPECS/ci-sssd.spec; \
```
alexey-tikhonov: Won't it strip %{?dist}?

src/rpms/Makefile
```diff
@@ -1,3 +1,13 @@
+set-version: prep
+	@dnf install -y git
+	@if git describe; then \
+		GIT_VERSION=$$(git describe | sed 's/-/_/g'); \
```
alexey-tikhonov, line 4: Does CI checkout repo with tags?

src/rpms/Makefile
```diff
@@ -1,3 +1,13 @@
+set-version: prep
+	@dnf install -y git
```
alexey-tikhonov, line 2: Maybe better to include in image?

alexey-tikhonov:
Shouldn't it also remove
https://github.com/SSSD/sssd-ci-containers/blob/abac5495e151842924f5698919c3fc227cbc8bf5/src/rpms/ci-sssd.spec#L1
?

src/rpms/Makefile
alexey-tikhonov: And isn't `--tags` required?

src/rpms/Makefile
justin-stephenson: You are right, fixed.

src/rpms/Makefile
justin-stephenson: COPR `make srpm` flow triggered by webhook (push to sssd-ci-containers master branch) does clone/checkout repo including tags.

I created an annotated tag in testing so `--tags` was not necessary.

src/rpms/Makefile
justin-stephenson: I tried adding this package to be installed in the chroot in the COPR WebUI, but it still failed without the `dnf install git` command in the makefile

image

justin-stephenson:
> Shouldn't it also remove
> 
> https://github.com/SSSD/sssd-ci-containers/blob/abac5495e151842924f5698919c3fc227cbc8bf5/src/rpms/ci-sssd.spec#L1
> 
> ?

Yes, but this is currently a fallback of `git describe` fails. Maybe it is not needed however?

alexey-tikhonov:
> > Shouldn't it also remove
> > https://github.com/SSSD/sssd-ci-containers/blob/abac5495e151842924f5698919c3fc227cbc8bf5/src/rpms/ci-sssd.spec#L1
> > 
> > ?
> 
> Yes, but this is currently a fallback of `git describe` fails. Maybe it is not needed however?

Imo, it's better to hard fail instead of fallback. It's known that `build_timestamp` causes issues, so what's the point...

But let's see what @pbrezina says.

pbrezina:
I also think we should not fallback since the fallback does not work. But if current version works I think we can finally merge it to fix the issue.

justin-stephenson:
> > > Shouldn't it also remove
> > > https://github.com/SSSD/sssd-ci-containers/blob/abac5495e151842924f5698919c3fc227cbc8bf5/src/rpms/ci-sssd.spec#L1
> > > 
> > > ?
> > 
> > 
> > Yes, but this is currently a fallback of `git describe` fails. Maybe it is not needed however?
> 
> Imo, it's better to hard fail instead of fallback. It's known that `build_timestamp` causes issues, so what's the point...
> 

This is now removed.

alexey-tikhonov (approved):

PR #181, opened 2026-05-27
ldap: fix certificate import on F45

pbrezina:
--nickname should be optional, but if ommitted, it throws an error

```
Error: expected str, bytes or os.PathLike object, not NoneType
```

See: https://github.com/389ds/389-ds-base/issues/7470

### DIFF
--- a/src/ansible/roles/ldap/tasks/main.yml
+++ b/src/ansible/roles/ldap/tasks/main.yml
@@ -16,7 +16,15 @@
   shell: |
     dsconf localhost security ca-certificate add --file /var/data/certs/ca.crt --name "sssd-ca"
     dsconf localhost security ca-certificate set-trust-flags "sssd-ca" --flags "CT,,"
-    dsctl localhost tls import-server-key-cert /var/data/certs/master.ldap.test.crt /var/data/certs/master.ldap.test.key
+
+    ARGS=
+    # Check if --nickname parameter is supported and set it accordingly
+    # Workaround for https://github.com/389ds/389-ds-base/issues/7470
+    if dsctl localhost tls import-server-key-cert --help 2>&1 | grep -q -- '--nickname'; then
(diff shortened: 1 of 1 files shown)


jakub-vavra-cz (changes requested):
The ci seems red for other versions.

I think the "correct" fix is to split the shell and have the faiing command with block+rescue or use this a s a fallback:

dsctl localhost tls import-server-key-cert /var/data/certs/master.ldap.test.crt /var/data/certs/master.ldap.test.key || dsctl localhost tls import-server-key-cert --nickname "Server-Cert" /var/data/certs/master.ldap.test.crt /var/data/certs/master.ldap.test.key

pbrezina:
Works now.

src/ansible/roles/ldap/tasks/main.yml
```diff
@@ -16,7 +16,15 @@
   shell: |
     dsconf localhost security ca-certificate add --file /var/data/certs/ca.crt --name "sssd-ca"
     dsconf localhost security ca-certificate set-trust-flags "sssd-ca" --flags "CT,,"
-    dsctl localhost tls import-server-key-cert /var/data/certs/master.ldap.test.crt /var/data/certs/master.ldap.test.key
+
+    ARGS=
+    # Check if --nickname parameter is supported and set it accordingly
+    # Workaround for https://github.com/389ds/389-ds-base/issues/7470
+    if dsctl localhost tls import-server-key-cert --help 2>&1 | grep -q -- '--nickname'; then
+      ARGS+='--nickname "Server-Cert"'
```
alexey-tikhonov, line 24: Shouldn't this be `ARGS+='--nickname Server-Cert'` (without "")?

jakub-vavra-cz (approved):

alexey-tikhonov (approved):

PR #182, opened 2026-06-04
ldap: Remove quotes from Server-Cert nickname

justin-stephenson:
Avoid 389ds creating a new server certificate with "" in the certificate nickname like "Server-Cert"


sumit-bose (approved):
Hi,

the fix is working well in my tests, so ACK.

Not sure if you want to address the bash-ism mentioned by Gemini?

bye,
Sumit

justin-stephenson:
> Hi,
> 
> the fix is working well in my tests, so ACK.
> 
> Not sure if you want to address the bash-ism mentioned by Gemini?
> 
> bye, Sumit

Thanks, I'll ignore it since we dropped support for Debian / Ubuntu some time ago.

PR #185, opened 2026-06-15
passkey: comment out starting vfido service

spoore1:
In some instances vfido is started in an environment where it does not
properly function.   This is a workaround to skip starting it during
playbook runs using the passkey role.


spoore1:
Closing this since the fix will be more invasive and require updates to sssd-test-framework as well.   Forcing "passkey_support: no" for the playbook_vm.yaml resolves the issue.

PR #189, opened 2026-06-17
Performance: add large ldap data

thalman:
This PR creates large content in LDAP server. This will be used by performance tests.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -327,3 +327,92 @@ are:
 
     base-keycloak --> keycloak
 ```
+
+## Pre-populated data
+
+For performance testing we have containers with prepopulated content.
+This content is used in some topologies (e.g. BigLDAP) to avoid
+creating large amount of users and groups during the test.
+
+### ldap container
+
+The ldap container includes two backup datasets: `base` and `largedata`.
+
+For your own experiments you can restore either backup using
+the `dsctl bak2db` command. The Directory Server **must be stopped**
+before performing the restore operation:
+
+    systemctl stop dirsrv@localhost.service
+    dsctl slapd-localhost bak2db largedata
+    systemctl start dirsrv@localhost.service
+
+This is done automatically by BigLDAP topology - no need to
+write extra steps in the test.
+
+### Users and groups in the containers
+
+In directory server we have POSIX-compatible users. All users use `posixAccount`
+and `inetUser` object classes with predefined password is `Secret123`.
+
+Groups use `posixGroup` and `groupOfNames` object classes. They have `memberUID`
+and `member` attributes to support both direct user membership (`rfc2307`) and
+nested groups (`rfc2307bis`).
+
+#### `group_1k_XY`
+
+We have 50 groups `group_1k_01`, `group_1k_02`, .. `group_1k_50`.
+
+Every group has 1000 users - `member_1k_<XY>_0001` to `member_1k_<XY>_1000`.
+
+#### `group_plain_jumbo`
+
+This is a large group containing directly 50000 users. Users in this group
+are called `member_plain_jumbo_00001` .. `member_plain_jumbo_50000`
+
+#### Nested groups
+
+Nested group name syntax is `nested_group_<index>_<depth>`.
+
+
+#### `group_1k_XY`
+
+We have 50 groups `group_1k_01`, `group_1k_02`, .. `group_1k_50`.
+
+Every group has 1000 users - `member_1k_<XY>_0001` to `member_1k_<XY>_1000`.
+
+#### `group_plain_jumbo`
+
+This is a large group containing directly 50000 users. Users in this group
+are called `member_plain_jumbo_00001` .. `member_plain_jumbo_50000`
+
+#### Nested groups
+
+Nested group name syntax is `nested_group_<index>_<depth>`.
+Groups and their content looks like
+
+    - nested_group_1_1
+        - group_1k_01
+        ...
+        - group_1k_10
+        - nested_group_1_2
+            - group_1k_11
+            ...
+            - group_1k_20
+            - nested_group_1_3
+                - group_1k_21
+                ...
+                - group_1k_30
+    - nested_group_2_1
+        - group_1k_31
+        ...
+        - group_1k_40
+        - nested_group_2_2
+            - group_1k_41
+            ...
+            - group_1k_50
+
+#### Users with many groups
+
+User `user_10k` is member of `group_1k_01` .. `group_1k_10`
+
+User `user_30k` is member of `group_1k_01` .. `group_1k_30`
--- a/readme.md
+++ b/readme.md
@@ -327,3 +327,76 @@ are:
 
     base-keycloak --> keycloak
 ```
+
+## Pre-populated data
+
+For performance testing we have containers with prepopulated content.
+This content is used in some topologies (e.g. BigLDAP) to avoid
+creating large amount of users and groups during the test.
+
+### ldap container
+
+The ldap container includes two backup datasets: `base` and `largedata`.
+
+For your own experiments you can restore either backup using
+the `dsctl bak2db` command. The Directory Server **must be stopped**
+before performing the restore operation:
+
+    systemctl stop dirsrv@localhost.service
+    dsctl slapd-localhost bak2db largedata
+    systemctl start dirsrv@localhost.service
+
+This is done automatically by BigLDAP topology - no need to
+write extra steps in the test.
+
+### Users and groups in the containers
+
+In directory server we have POSIX-compatible users. All users use `posixAccount`
+and `inetUser` object classes with a predefined password `Secret123`.
+
+Groups use `posixGroup` and `groupOfNames` object classes. They have `memberUID`
+and `member` attributes to support both direct user membership (`rfc2307`) and
+nested groups (`rfc2307bis`).
+
+#### `group_1k_XY`
+
+We have 50 groups `group_1k_01`, `group_1k_02`, .. `group_1k_50`.
+
+Every group has 1000 users - `member_1k_<XY>_0001` to `member_1k_<XY>_1000`.
+
+#### `group_plain_jumbo`
+
+This is a large group containing directly 50000 users. Users in this group
+are called `member_plain_jumbo_00001` .. `member_plain_jumbo_50000`
+
+#### Nested groups
+
+Nested group name syntax is `nested_group_<index>_<depth>`.
+Groups and their content looks like
+
+    - nested_group_1_1
+        - group_1k_01
+        ...
+        - group_1k_10
+        - nested_group_1_2
+            - group_1k_11
+            ...
+            - group_1k_20
+            - nested_group_1_3
+                - group_1k_21
+                ...
+                - group_1k_30
+    - nested_group_2_1
+        - group_1k_31
+        ...
+        - group_1k_40
+        - nested_group_2_2
+            - group_1k_41
+            ...
+            - group_1k_50
+
+#### Users with many groups
+
+User `user_10k` is member of `group_1k_01` .. `group_1k_10`
+
+User `user_30k` is member of `group_1k_01` .. `group_1k_30`
--- a/src/ansible/roles/ldap/files/generate-bigldap-data.py
+++ b/src/ansible/roles/ldap/files/generate-bigldap-data.py
@@ -1,0 +1,140 @@
+#!/usr/bin/env python
+
+from inspect import cleandoc
+
+
+class ldifCreator:
+    def __init__(self):
+        self.uid = 30000
+        self.gid = 230000
+
+    def write_base_structure(self):
+        """Write the base LDAP structure (domain, OUs)."""
+        print(
+            cleandoc(
+                """
+                dn: dc=ldap,dc=test
+                objectClass: top
+                objectClass: domain
+                dc: ldap
+                description: dc=ldap,dc=test
+                aci: (targetattr="dc || description || objectClass")(targetfilte
+                 r="(objectClass=domain)")(version 3.0; acl "Enable anyone domai
+                 n read"; allow (read, search, compare)(userdn="ldap:///anyone")
+                 ;)
+                aci: (targetattr=*)(version 3.0; acl "Enable anyone read"; allow
+                 (read, search, compare)(userdn="ldap:///anyone");)
+
+                dn: ou=users,dc=ldap,dc=test
+                objectClass: top
+                objectClass: organizationalUnit
+                ou: users
+
+                dn: ou=groups,dc=ldap,dc=test
+                objectClass: top
+                objectClass: organizationalUnit
+                ou: groups
+                """
+            )
+        )
+        print("")
+
+    def write_user(self, username):
+        print(
+            cleandoc(
+                f"""
+                dn: cn={username},ou=users,dc=ldap,dc=test
+                objectClass: posixAccount
+                objectClass: inetuser
+                objectClass: top
+                cn: {username}
+                uid: {username}
+                uidNumber: {self.uid}
+                gidNumber: {self.uid}
+                homeDirectory: /home/{username}
+                userPassword:: e1NIQTI1Nn1MdEJuWm5sZFdLVHlMVkVhWnk4Z3ByQ1cwLzViV
+                 nE4NmRFWjRxYU5XL1lJPQ==
+                """
+            )
+        )
+        print("")
+        self.uid += 1
+
+    def write_group(self, groupname, users, groups):
+        print(
+            cleandoc(
+                f"""
+                dn: cn={groupname},ou=groups,dc=ldap,dc=test
+                objectClass: posixGroup
+                objectClass: groupOfNames
+                objectClass: top
+                cn: {groupname}
+                gidNumber: {self.gid}
+                """
+            )
+        )
+        if users:
+            for member in users:
+                print(f"member: cn={member},ou=users,dc=ldap,dc=test")
+                print(f"memberUID: {member}")
+        if groups:
+            for member in groups:
+                print(f"member: cn={member},ou=groups,dc=ldap,dc=test")
+        print("")
+        self.gid += 1
+
+    def write_user_10k_30k(self):
+        self.write_user("user_10k")
+        self.write_user("user_30k")
+
+    def write_group_1k_XY(self):
+        for group in range(1, 51):
+            members = []
+            group_name = "group_1k_%.2d" % group
+            for user in range(1, 1001):
+                user_name = "member_1k_%.2d_%.4d" % (group, user)
+                members.append(user_name)
+                self.write_user(user_name)
+            if group >= 1 and group <= 10:
+                members.append("user_10k")
+            if group >= 1 and group <= 30:
+                members.append("user_30k")
+            self.write_group(group_name, members, None)
+
+    def write_group_plain_jumbo(self):
+        members = []
+        for user in range(1, 50001):
+            user_name = "member_plain_jumbo_%.5d" % (user)
+            members.append(user_name)
+            self.write_user(user_name)
+        self.write_group("group_plain_jumbo", members, None)
+
+    def _group_1k_list(self, start, end):
+        members = []
+        for a in range(start, end + 1):
+            members.append("group_1k_%.2d" % (a))
+        return members
+
+    def write_nested_groups(self):
+        self.write_group(
+            "nested_group_1_1", None, self._group_1k_list(1, 10) + ["nested_group_1_2"]
+        )
+        self.write_group(
+            "nested_group_1_2", None, self._group_1k_list(11, 20) + ["nested_group_1_3"]
+        )
+        self.write_group("nested_group_1_3", None, self._group_1k_list(21, 30))
+        self.write_group(
+            "nested_group_2_1", None, self._group_1k_list(31, 40) + ["nested_group_2_2"]
+        )
+        self.write_group("nested_group_2_2", None, self._group_1k_list(41, 50))
+
+    def run(self):
+        self.write_base_structure()
+        self.write_group_plain_jumbo()
+        self.write_user_10k_30k()
+        self.write_group_1k_XY()
+        self.write_nested_groups()
+
+
+if __name__ == "__main__":
+    ldifCreator().run()
--- a/src/ansible/roles/ldap/files/generate-bigldap-data.py
+++ b/src/ansible/roles/ldap/files/generate-bigldap-data.py
@@ -1,0 +1,154 @@
+#!/usr/bin/env python3
+
+from inspect import cleandoc
+
+
+class LdifCreator:
+    def __init__(self):
+        self.uid = 30000
+        self.gid = 230000
+
+    def write_base_structure(self):
+        print(
+            cleandoc(
+                """
+                dn: dc=ldap,dc=test
+                objectClass: top
+                objectClass: domain
+                dc: ldap
+                description: dc=ldap,dc=test
+                aci: (targetattr="dc || description || objectClass")(targetfilte
+                 r="(objectClass=domain)")(version 3.0; acl "Enable anyone domai
+                 n read"; allow (read, search, compare)(userdn="ldap:///anyone")
+                 ;)
+                aci: (targetattr=*)(version 3.0; acl "Enable anyone read"; allow
+                 (read, search, compare)(userdn="ldap:///anyone");)
+
+                dn: ou=users,dc=ldap,dc=test
+                objectClass: top
+                objectClass: organizationalUnit
+                ou: users
+
+                dn: ou=groups,dc=ldap,dc=test
+                objectClass: top
+                objectClass: organizationalUnit
+                ou: groups
+                """
+            )
+        )
+        print("")
+
+    def write_user(self, username):
+        print(
+            cleandoc(
+                f"""
+                dn: cn={username},ou=users,dc=ldap,dc=test
+                objectClass: posixAccount
+                objectClass: inetUser
+                objectClass: top
+                cn: {username}
+                uid: {username}
+                uidNumber: {self.uid}
+                gidNumber: {self.uid}
+                homeDirectory: /home/{username}
+                userPassword:: e1NIQTI1Nn1MdEJuWm5sZFdLVHlMVkVhWnk4Z3ByQ1cwLzViV
+                 nE4NmRFWjRxYU5XL1lJPQ==
+                """
+            )
+        )
+        print("")
+        print(
+            cleandoc(
+                f"""
+                dn: cn={username},ou=groups,dc=ldap,dc=test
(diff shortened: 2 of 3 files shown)


src/ansible/roles/ldap/tasks/main.yml
```diff
@@ -44,6 +44,25 @@
...
+- name: Create big ldap data binary backup
+  shell: |
+    set -e
+    systemctl stop dirsrv@localhost.service || true
+    dsctl slapd-localhost db2bak base
+    dsctl slapd-localhost ldif2db userroot /var/lib/dirsrv/slapd-localhost/bak/largedata.ldif
+    dsctl slapd-localhost db2bak largedata
+    dsctl slapd-localhost bak2db base
+  args:
+    creates: '/var/lib/dirsrv/slapd-localhost/bak/largedata'
```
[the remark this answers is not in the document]
thalman: gemini is wrong, dsctl will not fail if the target folder exists.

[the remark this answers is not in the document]
jakub-vavra-cz: I would rather see this split in multiple steps:
1. Delete old and create "base" backup.
2. In block load and dump large data
3. In always section of the 2. block restore the base

alexey-tikhonov:
@thalman, IIUC, this populates our base LDAP CI container? Wouldn't it be better to have a new - "ldap-populated" - container to avoid potential inference with tests that expect empty LDAP?

Another note: please also add a brief description of the content (a 'readme' somewhere?)

src/ansible/roles/ldap/tasks/main.yml
```diff
@@ -52,6 +44,25 @@
       add: objectclasses
       objectclasses: ( 2.16.840.1.113730.3.8.24.9 NAME 'passkeyUser' DESC 'IPA passkey user' AUXILIARY MAY passkey)
 
+- name: Copy big ldif to backup folder
+  copy:
+    src: largedata.ldif
+    dest: /var/lib/dirsrv/slapd-localhost/bak
```
jakub-vavra-cz, line 50: Add a step that check is the directory exists and creates it if not.

jakub-vavra-cz:
The ldiff is ~ 81MB and I am not sure how big becomes the binary dump.
It might be better not to carry those extras to basic ldap, or at least add creation/handling of the
binary dump directly to the topology setup for a new topo like LDAPLarge that will need it.

alexey-tikhonov:
> or at least add creation/handling of the
> binary dump directly to the topology setup for a new topo like LDAPLarge that will need it

The idea was to have it handy not only for automated tests in PR CI but also for "manual" experiments. So I'd vote for a separate container image. If it's compatible with test-framework.

thalman:
> @thalman, IIUC, this populates our base LDAP CI container? Wouldn't it be better to have a new - "ldap-populated" - container to avoid potential inference with tests that expect empty LDAP?
> 
> Another note: please also add a brief description of the content (a 'readme' somewhere?)

I was thinking about separate container but considering resources and simplicity I decided for this approach. It keeps the ldap empty, just creates backup of large data.

BigLDAP topology recovers them (restore is quick) and framework itself take care of restoring "empty" state.

README - yes there should be one but I think that better place would be in SSSD project in `src/tests/performance` - because this is where I plan to store performance tests. What do you think?

thalman:
> The ldiff is ~ 81MB and I am not sure how big becomes the binary dump. It might be better not to carry those extras to basic ldap, or at least add creation/handling of the binary dump directly to the topology setup for a new topo like LDAPLarge that will need it.

Regarding the resources - having two containers would consume more disk and more memory.

Binary backup is about 270,000K

All these steps here are done for new topology to be able quicky restore large data (see https://github.com/SSSD/sssd-test-framework/pull/251) and it does not affect our current LDAP topology/tests.

src/ansible/roles/ldap/tasks/main.yml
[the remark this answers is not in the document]
thalman: Why to delete old "base" backup? I'm not doing that. Do you think that I should do what gemini suggests? --`rm -rf /var/lib/dirsrv/slapd-localhost/bak/base`

src/ansible/roles/ldap/tasks/main.yml
thalman: Why? copy creates whole path and /var/lib/dirsrv/slapd-localhost is created just in previous step (creating the dirsrv instance).

thalman:
@alexey-tikhonov, @jakub-vavra-cz - github comlains that the ldif is too big.

`Large files detected. You may want to try Git Large File Storage - https://git-lfs.github.com `

Do you think it would be good idea to `gzip` it? it will be just 5.000k but binary. This file should not change much.

src/ansible/roles/ldap/tasks/main.yml
thalman: I updated the role. Please take a look.

jakub-vavra-cz:
I would drop the ldif file after having the binary dump and ignore the gzip or  Git Large File Storage .

alexey-tikhonov:
> README - yes there should be one but I think that better place would be in SSSD project in `src/tests/performance` - because this is where I plan to store performance tests. What do you think?

I think description of the content should be in the same repo as the content itself, i.e. here.
It might be a separate file made available in the container file system.

jakub-vavra-cz (approved):
LGTM

thalman:
I extended the readme once more describing users and groups that we have.

But DO NOT MERGE! I found one group in ldif that is empty by mistake. Let me fix that before merging

thalman:
> I extended the readme once more describing users and groups that we have.
> 
> But DO NOT MERGE! I found one group in ldif that is empty by mistake. Let me fix that before merging

Group is fixed.

Readme extedend of data description.

readme.md
```diff
@@ -327,3 +313,67 @@ are:
...
+        - nested_group_2_2
+            - group_1k_41
+            ...
+            - group_1k_50
+
+#### Users with many groups
+
+User `user_10k` is member of `group_1k_01` .. `group_1k_10`
+
+User `user_30x` is member of `group_1k_01` .. `group_1k_30`
```
alexey-tikhonov, line 378: `user_30x` or `user_30k`?

readme.md
```diff
@@ -327,3 +313,67 @@ are:
 
     base-keycloak --> keycloak
 ```
+
+## Pre-populated data
+
+For performance testing we have containers with prepopulated content.
+This content is used in some topologies (e.g. BigLDAP) to avoid
```
alexey-tikhonov, line 320: Please mention what LDAP schema is used.

readme.md
```diff
@@ -327,3 +313,67 @@ are:
...
+
+The ldap container includes two backup datasets: `base` and `largedata`.
+
+For your own experiments you can restore either backup using
+the `dsctl bak2db` command. The Directory Server **must be stopped**
+before performing the restore operation:
+
+    systemctl stop dirsrv@localhost.service
+    dsctl slapd-localhost bak2db largedata
+    systemctl start dirsrv@localhost.service
```
alexey-tikhonov, line 333: Should this work with 'sssd-ldap' container "out of the box" or is some configuration needed?

IIRC, in my tests I had to extend some limits in ds config to serve large groups...

alexey-tikhonov:
Initially I thought we were going to have a dedicated container image with DS pre-populated, but it was decided to go another way - that's fine.

But I don't think that having '.ldif' file of "1,620,637 Lines" in the repo makes a lot of sense.
Would it be better to store a Python script that generates .ldif?

@pbrezina, what would you say?

thalman:
@alexey-tikhonov I had to redo the script for generating LDAP data, but here it is. To keep data consistent I decided to extend ansible role a bit, mainly running memberof fixup task to ensure consistency and simplify generation script.

readme.md
thalman: Good catch, typo fixed

readme.md
thalman: Readme extended

readme.md
thalman: Yes, you do not have to do that. What I mean here is how to get the data live if you want to check them or play with them. I added one sentence into readme to clarify that this is done automatically in the test.

thalman: Regarding the limits - I had issue when using `ldapmodify` for filling the DS. I do not see the problem when using `dsctl ldif2db`. I did not have issue reading large groups - I just double-checked that with the jumbo group. Works for me in default state.

Just for the record, this is the setting needed for `ldapadd/ldapmodify`

```
dn: cn=config
changetype: modify
replace: nsslapd-sizelimit
nsslapd-sizelimit: 15000

dn: cn=config,cn=ldbm database,cn=plugins,cn=config
changetype: modify
replace: nsslapd-lookthroughlimit
nsslapd-lookthroughlimit: 10000
```

alexey-tikhonov: > nsslapd-sizelimit: 15000
> nsslapd-lookthroughlimit: 10000

That exactly what I meant. Good to know it's needed for `ldapadd`/`ldapmodify` only.

alexey-tikhonov:
@jakub-vavra-cz, PR was substantially updated (re-written), please review again.

src/ansible/roles/ldap/files/generate-bigldap-data.py
```diff
@@ -0,0 +1,140 @@
+#!/usr/bin/env python
+
+from inspect import cleandoc
+
+
+class ldifCreator:
```
jakub-vavra-cz: ldifCreator should be LdifCreator (PEP 8).

readme.md
```diff
@@ -327,3 +327,92 @@ are:
...
+Every group has 1000 users - `member_1k_<XY>_0001` to `member_1k_<XY>_1000`.
+
+#### `group_plain_jumbo`
+
+This is a large group containing directly 50000 users. Users in this group
+are called `member_plain_jumbo_00001` .. `member_plain_jumbo_50000`
+
+#### Nested groups
+
+Nested group name syntax is `nested_group_<index>_<depth>`.
```
jakub-vavra-cz, line 374: Duplicated text?

readme.md
```diff
@@ -327,3 +327,92 @@ are:
...
+    dsctl slapd-localhost bak2db largedata
+    systemctl start dirsrv@localhost.service
+
+This is done automatically by BigLDAP topology - no need to
+write extra steps in the test.
+
+### Users and groups in the containers
+
+In directory server we have POSIX-compatible users. All users use `posixAccount`
+and `inetUser` object classes with predefined password is `Secret123`.
```
jakub-vavra-cz:
```suggestion
In directory server we have POSIX-compatible users. All users use `posixAccount`
and `inetUser` object classes with a predefined password `Secret123`.
```

src/ansible/roles/ldap/files/generate-bigldap-data.py
```diff
@@ -0,0 +1,140 @@
...
+            members = []
+            group_name = "group_1k_%.2d" % group
+            for user in range(1, 1001):
+                user_name = "member_1k_%.2d_%.4d" % (group, user)
+                members.append(user_name)
+                self.write_user(user_name)
+            if group >= 1 and group <= 10:
+                members.append("user_10k")
+            if group >= 1 and group <= 30:
+                members.append("user_30k")
```
jakub-vavra-cz: group is always >=1 because of the range

src/ansible/roles/ldap/files/generate-bigldap-data.py
```diff
@@ -0,0 +1,140 @@
...
+        )
+        print("")
+
+    def write_user(self, username):
+        print(
+            cleandoc(
+                f"""
+                dn: cn={username},ou=users,dc=ldap,dc=test
+                objectClass: posixAccount
+                objectClass: inetuser
```
jakub-vavra-cz: Should there be inetUser instead as documented in readme?

src/ansible/roles/ldap/tasks/main.yml
```diff
@@ -52,6 +52,49 @@
...
+  - name: Create large ldap binary backup
+    shell: |
+      /usr/local/bin/generate-bigldap-data.py > /var/lib/dirsrv/slapd-localhost/bak/largedata.ldif && \
+      dsctl slapd-localhost ldif2db userroot /var/lib/dirsrv/slapd-localhost/bak/largedata.ldif && \
+      systemctl start dirsrv@localhost.service && \
+      dsconf slapd-localhost plugin memberof enable && \
+      systemctl restart dirsrv@localhost.service && \
+      dsconf slapd-localhost plugin memberof fixup --wait dc=ldap,dc=test && \
+      systemctl stop dirsrv@localhost.service && \
+      dsctl slapd-localhost db2bak largedata
```
jakub-vavra-cz: I would rather see this split to more standalone commands so failure can be easily troubleshooted especially with commands that might run for a while like the fixup.

src/ansible/roles/ldap/tasks/main.yml
```diff
@@ -52,6 +52,49 @@
...
+- name: Create base ldap binary backup
+  shell: |
+    dsctl slapd-localhost db2bak base
+  args:
+    creates: '/var/lib/dirsrv/slapd-localhost/bak/base'
+
+- name: 'Install BigLDAP generator'
+  copy:
+    src: generate-bigldap-data.py
+    dest: /usr/local/bin/generate-bigldap-data.py
```
jakub-vavra-cz: The file will not be needed in the container aftewards, You might generate the ldif on the executor, and push it to the container instead.

src/ansible/roles/ldap/files/generate-bigldap-data.py
```diff
@@ -0,0 +1,140 @@
...
+from inspect import cleandoc
+
+
+class ldifCreator:
+    def __init__(self):
+        self.uid = 30000
+        self.gid = 230000
+
+    def write_base_structure(self):
+        """Write the base LDAP structure (domain, OUs)."""
```
jakub-vavra-cz: Either add proper docstrings everywhere or drop this one for consistency.

jakub-vavra-cz (changes requested):

src/ansible/roles/ldap/files/generate-bigldap-data.py
thalman: well this is case insensitive so for the sake of readability I use "inetUser" in readme but the schema in DS server present this in lowercase.

src/ansible/roles/ldap/files/generate-bigldap-data.py
thalman: fixed

readme.md
thalman: Sorry for the mess, fixed

readme.md
thalman: fixed

src/ansible/roles/ldap/files/generate-bigldap-data.py
thalman: fixed

src/ansible/roles/ldap/files/generate-bigldap-data.py
thalman: fixed

src/ansible/roles/ldap/tasks/main.yml
thalman: splitted, JFI fixup task takes ~10s on my laptop.

src/ansible/roles/ldap/tasks/main.yml
thalman: Because the output is huge, this is not recommended due to ansible memory consumption (at least that is what I found).

So if you do not mind I would keep it that way. But I added "always:" task that removes unnecessary ldif + generator

src/ansible/roles/ldap/files/generate-bigldap-data.py
thalman: doc string dropped

src/ansible/roles/ldap/tasks/main.yml
```diff
@@ -52,6 +52,72 @@
...
       objectclasses: ( 2.16.840.1.113730.3.8.24.9 NAME 'passkeyUser' DESC 'IPA passkey user' AUXILIARY MAY passkey)
 
+- name: Stop LDAP service
+  service:
+    name: dirsrv@localhost.service
+    state: stopped
+
+- name: Create base ldap binary backup
+  shell: |
+    dsctl slapd-localhost db2bak base
```
jakub-vavra-cz: command:

src/ansible/roles/ldap/tasks/main.yml
```diff
@@ -52,6 +52,72 @@
...
+      src: generate-bigldap-data.py
+      dest: /usr/local/bin/generate-bigldap-data.py
+      owner: root
+      group: root
+      mode: '0755'
+  - name: Create large data ldif
+    shell: |
+      /usr/local/bin/generate-bigldap-data.py > /var/lib/dirsrv/slapd-localhost/bak/largedata.ldif
+  - name: Create large ldap binary backup
+    shell: |
```
jakub-vavra-cz: command:

src/ansible/roles/ldap/tasks/main.yml
```diff
@@ -52,6 +52,72 @@
...
+      dsctl slapd-localhost ldif2db userroot /var/lib/dirsrv/slapd-localhost/bak/largedata.ldif
+  - name: Start LDAP service
+    service:
+      name: dirsrv@localhost.service
+      state: started
+  - name: Enable memberof plugin
+    shell:
```
jakub-vavra-cz: command:

src/ansible/roles/ldap/tasks/main.yml
```diff
@@ -52,6 +52,72 @@
...
+      dsconf slapd-localhost plugin memberof enable
+  - name: Restart LDAP service
+    service:
+      name: dirsrv@localhost.service
+      state: restarted
+  - name: Run DS fixup task
+    shell:
```
jakub-vavra-cz: command:

src/ansible/roles/ldap/tasks/main.yml
```diff
@@ -52,6 +52,72 @@
...
+      dsconf slapd-localhost plugin memberof fixup --wait dc=ldap,dc=test
+  - name: Stop LDAP service
+    service:
+      name: dirsrv@localhost.service
+      state: stopped
+  - name: Create large data backup
+    shell:
```
jakub-vavra-cz: command:

src/ansible/roles/ldap/tasks/main.yml
```diff
@@ -52,6 +52,72 @@
...
+      dsctl slapd-localhost db2bak largedata
+  when: not largedata_backup.stat.exists
+  always:
+  - name: Restore base backup
+    shell: |
```
jakub-vavra-cz: command:

jakub-vavra-cz (changes requested):
Change shell: to command: where possible. (keeping "Create large data ldif")

src/ansible/roles/ldap/tasks/main.yml
thalman: done

src/ansible/roles/ldap/tasks/main.yml
thalman: done

src/ansible/roles/ldap/tasks/main.yml
thalman: done

src/ansible/roles/ldap/tasks/main.yml
thalman: done

src/ansible/roles/ldap/tasks/main.yml
thalman: done

src/ansible/roles/ldap/tasks/main.yml
thalman: done

jakub-vavra-cz (approved):

src/ansible/roles/ldap/tasks/main.yml
```diff
@@ -52,6 +52,68 @@
...
+      /usr/local/bin/generate-bigldap-data.py > /var/lib/dirsrv/slapd-localhost/bak/largedata.ldif
+  - name: Create large ldap binary backup
+    command: dsctl slapd-localhost ldif2db userroot /var/lib/dirsrv/slapd-localhost/bak/largedata.ldif
+  - name: Start LDAP service
+    service:
+      name: dirsrv@localhost.service
+      state: started
+  - name: Enable memberof plugin
+    command:
+      dsconf slapd-localhost plugin memberof enable
```
alexey-tikhonov: Why do we need server side memberof plugin?

src/ansible/roles/ldap/tasks/main.yml
```diff
@@ -52,6 +52,68 @@
...
+  - name: Stop LDAP service
+    service:
+      name: dirsrv@localhost.service
+      state: stopped
+  - name: Create large data backup
+    command: dsctl slapd-localhost db2bak largedata
+  when: not largedata_backup.stat.exists
+  always:
+  - name: Restore base backup
+    command: dsctl slapd-localhost bak2db base
```
alexey-tikhonov, line 89: If some of above steps fail, it can get here with ds running.

src/ansible/roles/ldap/files/generate-bigldap-data.py
```diff
@@ -0,0 +1,139 @@
+#!/usr/bin/env python
```
alexey-tikhonov: `python3`?

src/ansible/roles/ldap/tasks/main.yml
alexey-tikhonov: Moreover, this doesn't disable 'memberof' plugin back.
But I think plugin isn't needed at all - see above.

src/ansible/roles/ldap/tasks/main.yml
```diff
@@ -52,6 +52,68 @@
...
+  - name: Delete unnecessary files
+    file:
+      state: absent
+      path: "{{ item }}"
+      state: absent
```
alexey-tikhonov, line 93: `state: absent` is duplicated.

src/ansible/roles/ldap/files/generate-bigldap-data.py
```diff
@@ -0,0 +1,139 @@
...
+            cleandoc(
+                f"""
+                dn: cn={username},ou=users,dc=ldap,dc=test
+                objectClass: posixAccount
+                objectClass: inetUser
+                objectClass: top
+                cn: {username}
+                uid: {username}
+                uidNumber: {self.uid}
+                gidNumber: {self.uid}
```
alexey-tikhonov, line 52: Are group objects with those GIDs created?

src/ansible/roles/ldap/tasks/main.yml
thalman: We do not need it for tests, but in my experience such plugins are used very often and so having it enabled makes the environment closer to real world (but I do not have strong data on this).

I also leverage the capability to create "memberOf" attribute for me.

If you think that we need it off, let me know.

src/ansible/roles/ldap/files/generate-bigldap-data.py
thalman: fixed

src/ansible/roles/ldap/tasks/main.yml
thalman: fixed

src/ansible/roles/ldap/files/generate-bigldap-data.py
thalman: Private groups are not created. Can be done easily if we want them.

Should I extend the generator?

src/ansible/roles/ldap/tasks/main.yml
thalman: stopping DS fixed. For member of plugin see the discussion above.

src/ansible/roles/ldap/tasks/main.yml
alexey-tikhonov: > I also leverage the capability to create "memberOf" attribute for me.

But, IIRC, SSSD LDAP doesn't use it (maybe stores as `SYSDB_ORIG_MEMBEROF` but doesn't use).
Maybe IPA provider use it (instead of AD token-groups), but not LDAP.
@sumit-bose, please correct me if I'm wrong.

src/ansible/roles/ldap/files/generate-bigldap-data.py
alexey-tikhonov: I think so. Or maybe other existing group(s) could be used as primary group.

src/ansible/roles/ldap/tasks/main.yml
thalman: ok, I will switch it off after fixup task

alexey-tikhonov: Why do we need it at all?

src/ansible/roles/ldap/files/generate-bigldap-data.py
thalman: done

src/ansible/roles/ldap/tasks/main.yml
thalman: In the meeting we agreed that we do not want `memberOf` attribute in this topology. So I deleted the steps with `memberof` plugin and its' `fixup` task.

thalman:
@jakub-vavra-cz - based on Alexey's comment I removed steps around `memberof` plugin and added private user groups in ldif generator. Do you want to re-review?

src/ansible/roles/ldap/files/generate-bigldap-data.py
```diff
@@ -0,0 +1,152 @@
...
+        print("")
+        print(
+            cleandoc(
+                f"""
+                dn: cn={username},ou=groups,dc=ldap,dc=test
+                objectClass: posixGroup
+                objectClass: top
+                cn: {username}
+                gidNumber: {self.uid}
+                memberUID: {username}
```
alexey-tikhonov, line 69: Also `groupOfNames` and `member`?
At least to be consistent with `write_group()`

thalman: Well private groups are a bit special, I wanted them to be pure posix group. But why not, here it is.

jakub-vavra-cz (approved):
LGTM. Rebase seems to be needed.

thalman:
rebased once again

alexey-tikhonov (approved):

alexey-tikhonov:
Ok, let's merge this and try to actually build based on this, then we'll see if any tweaks will be required to the content.
ACK.

PR #190, opened 2026-06-17
passkey vfido.service remove - from ExecStartPre

spoore1:
The vfido.service unit file had ExecStartPre=+- to allow the modprobe to
fail silently.   This was causing a log file to fill because it could
not find the module in some setups.   Removing the - so that the
modprobe will cause the service to fail if it fails.


danlavu (approved):
Looks good to me

PR #193, opened 2026-07-01
Passkey: Fix vfido resource limits by pre-compiling binary

aborah-sudo:
Replace `go run` with pre-compiled binary to avoid Go compiler fork limits in Testing Farm containers. This eliminates the intermittent "resource temporarily unavailable" errors that caused ~3-4% passkey test failure rate.

### DIFF
--- a/src/ansible/roles/passkey/tasks/main.yml
+++ b/src/ansible/roles/passkey/tasks/main.yml
@@ -44,6 +44,21 @@
     with_fileglob:
       - "vfido_*"
 
+  - name: Clone virtual-fido to test_venv
+    git:
+      repo: "https://github.com/bulwarkid/virtual-fido.git"
+      dest: "{{ test_venv }}/virtual-fido"
+      version: 512d8a3fef0e073171a7ef3261f300e8a5e48694
+      clone: yes
+      update: yes
+      depth: 1
+
+  - name: Build vfido-demo binary once to avoid Go compiler fork limits
+    command: go build -ldflags="-s -w" -o {{ test_venv }}/bin/vfido-demo ./cmd/demo
+    args:
+      chdir: "{{ test_venv }}/virtual-fido"
+      creates: "{{ test_venv }}/bin/vfido-demo"
+
   - name: Install vfido systemd service unit file
     copy:
       src: vfido.service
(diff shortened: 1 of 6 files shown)


aborah-sudo:
The build is using Python 3.15 beta (Fedora 45 rawhide), and there's an incompatibility with Ansible:

  python3                    x86_64 0:3.15.0~b3-1.fc45
  AttributeError: module 'dataclasses' has no attribute '_is_type'

src/ansible/roles/passkey/tasks/main.yml
```diff
@@ -65,4 +65,10 @@
       update: yes
       depth: 1
 
+  - name: Build vfido-demo binary once to avoid Go compiler fork limits
+    command: go build -o {{ test_venv }}/bin/vfido-demo ./cmd/demo
```
ikerexxe: Add also `-ldflags=-s -w` flags. This will help create a smaller binary that loads faster and has better memory efficiency

ikerexxe (approved):
LGTM!

spoore1 (changes requested):
While this generally looks good, I'm seeing test failures when I manually run the GDM passkey tests.   I'm able to reproduce the issue with manual steps:

```
[root@client ~]# /opt/test_venv/bin/vfido_reset
[root@client ~]# /opt/test_venv/bin/vfido_pin_enable
[LOG] Log Level Set: 2
PIN enabled
[root@client ~]# /opt/test_venv/bin/vfido_pin_set 123456
[LOG] Log Level Set: 2
PIN set
[root@client ~]# systemctl start vfido
[root@client ~]# sssctl passkey-register --username pkgdmuser1 --domain ldap.test
Please touch the device.
```

It should prompt for a PIN first.   I think this may need closer investigation or let me know if I need to test it with other sssd and sssd-test-framework PR/branches.

spoore1 (approved):
Looks like it's fixed with the latest updates to fix the current dir before running the commands:
```
2026-07-02T16:04:18 PASSED ../sssd/src/tests/system/tests/test_gdm.py::test_gdm__passkey_login_succeeds_with_pin (bareipa)
```

PR #198, opened 2026-07-17
Fix pytest-mh test environment for RHEL 8.4

mmatsuya:
## Summary

Two fixes to allow pytest-mh tests to run on RHEL 8.4 AWS images.

### 1. packages: skip shadow-utils-subid on RHEL < 8.6

`shadow-utils-subid` was introduced in RHEL 8.6 and is not available on
RHEL 8.4. Split it out from the minimal package set into a separate task
with a version condition so that RHEL 8.4 environments can still install
the rest of the minimal package set successfully.

### 2. common: rename /data to /data.bak if it already exists as a directory

RHEL 8.4 AWS images ship with `/data` as a pre-existing directory.
The `common` role attempts to create a backward compatibility symlink
`/data -> /var/data`, which fails when `/data` is already a directory.
Rename it to `/data.bak` first so the symlink can be created safely.
If `/data` does not exist (as on newer RHEL versions), this is a no-op.

## Test plan

- [x] Verify prep phase succeeds on RHEL 8.4
- [x] Verify tests run on RHEL 8.4 (120 passed, 14 failed due to RHEL 8.4 feature limitations)
- [x] Verify `shadow-utils-subid` is still installed on RHEL 8.6+

### DIFF
--- a/src/ansible/roles/packages/tasks/Fedora.yml
+++ b/src/ansible/roles/packages/tasks/Fedora.yml
@@ -68,11 +68,20 @@
       - policycoreutils
       - policycoreutils-python-utils
       - python3-pip
-      - shadow-utils-subid
       - sudo
       - rsync
       - passwd
       disable_gpg_check: yes
+
+  - name: Install shadow-utils-subid (RHEL 8.6+, all other distros)
+    dnf:
+      state: present
+      name:
+      - shadow-utils-subid
+      disable_gpg_check: yes
+    when: not (ansible_facts['distribution'] == 'RedHat' and
+               ansible_facts['distribution_major_version'] == '8' and
+               ansible_facts['distribution_version'] is version('8.6', '<'))
   when: "'base_ground' in group_names"
 
 - name: Install extended set of packages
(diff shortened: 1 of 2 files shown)


alexey-tikhonov (approved):

src/ansible/roles/packages/tasks/Fedora.yml
```diff
@@ -68,11 +68,20 @@
...
+
+  - name: Install shadow-utils-subid (RHEL 8.6+, all other distros)
+    dnf:
+      state: present
+      name:
+      - shadow-utils-subid
+      disable_gpg_check: yes
+    when: not (ansible_distribution == 'RedHat' and
+               ansible_distribution_major_version == '8' and
+               ansible_distribution_version is version('8.6', '<'))
```
jakub-vavra-cz: Please rewrite to the new style:
https://docs.ansible.com/projects/ansible/latest/porting_guides/porting_guide_13.html#inject-facts-as-vars

jakub-vavra-cz (changes requested):
Please fix the "ansible_" use so we do not have to rewrite it in future.

src/ansible/roles/packages/tasks/Fedora.yml
mmatsuya: Hi @jakub-vavra-cz, I updated the part to follow the new style. Thanks!

jakub-vavra-cz (approved):

PR #200, opened 2026-08-05
Fix STIG tier2: ignore passkey setup errors on hardened systems

madhuriupadhye:
STIG tier2 jobs fail during Ansible setup because passkey-related packages (umockdev, usbip) are not GPG-signed, and STIG-hardened systems enforce gpgcheck=1. This blocks the entire job even though STIG jobs do not run passkey tests (-k="not passkey and not smartcard").

Example error:

Failed to validate GPG signature for umockdev-0.19.1-2.el10.x86_64: Added ignore_errors: yes to passkey package install tasks in Fedora.yml and RedHat10.yml. This allows the setup to continue when passkey packages fail to install, unblocking STIG and other jobs that do not need them.

The passkey role's Build random.so task then fails with gcc: command not found on STIG systems because the earlier package install (which includes gcc) was silently ignored. The vfido wrapper block would similarly fail due to missing go. CentOS10.yml passkey tasks were also missing ignore_errors.

Added ignore_errors: yes to:
  - Both blocks in passkey/tasks/main.yml (build random.so, configure vfido)
  - Both passkey tasks in packages/tasks/CentOS10.yml (EPEL, usbip packages)

### DIFF
--- a/src/ansible/roles/packages/tasks/CentOS10.yml
+++ b/src/ansible/roles/packages/tasks/CentOS10.yml
@@ -53,6 +53,7 @@
   when:
     - passkey_support
     - "'base_client' in group_names or 'client' in group_names"
+  ignore_errors: yes
 
 - name: Install packages for usbip needed for virtual passkey testing
   dnf:
@@ -68,3 +69,4 @@
   when:
     - passkey_support
     - "'base_client' in group_names or 'client' in group_names"
+  ignore_errors: yes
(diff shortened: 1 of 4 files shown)


spoore1:
Since the ignore_errors could potentially cause issues with other environments where packages are missing, we may want to consider adding a variable for `stig_enabled` and using `failed_when` with something like this:
```
- name: test setup step
  dnf:
    name: package1
    state: present
  register: check_task
  failed_when:
    - check_task.rc != 0
    - stig_enabled != True
```

The `stig_enabled` option could be set to False as a default like `passkey_support` and added to `playbook_vm.yaml` the same way.   Then we can set as needed in other metadata when running `playbook_vm.yaml`

madhuriupadhye:
I created the new PR mistakenly as suggeseted by you, https://github.com/SSSD/sssd-ci-containers/pull/202.
marking this closed.

PR #201, opened 2026-08-05
Enable AES Kerberos encryption on AD server for FIPS:STIG client compatibility

madhuriupadhye:
- Enable AES-128 and AES-256 Kerberos encryption types on Windows AD domain controllers after forest creation
- Sets msDS-SupportedEncryptionTypes on DC computer objects and configures the registry-level Kerberos encryption policy
- Fixes AD STIG test failures where realm join fails with "The encryption types desired are not available in active directory" when the client has FIPS:STIG crypto policy applied

Problem
RHEL clients with FIPS:STIG crypto policy restrict Kerberos to strong AES encryption types only.
The Windows AD server in the CI test environment did not have AES enabled, causing all encryption types (18, 17, 16, 23, 3, 1) to be rejected during realm join. This blocked the entire tier2-stig-pytest-ad-stig test suite.

### DIFF
--- a/src/ansible/roles/ad/tasks/install.yml
+++ b/src/ansible/roles/ad/tasks/install.yml
@@ -30,6 +30,26 @@
   win_reboot:
   when: installation.changed
 
+- name: Enable AES Kerberos encryption for FIPS/STIG client compatibility
+  win_shell: |
+    Import-Module ActiveDirectory
+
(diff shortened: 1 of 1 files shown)


madhuriupadhye:
Closing this as multihost tests are failing and needs to add in multihost conftest.py.

PR #202, opened 2026-08-11
Add stig_enabled flag

madhuriupadhye:
- Add stig_enabled variable to playbook_vm.yml following the same pattern as passkey_support, defaults to false Pass override_stig_enabled: true as extra_vars on the playbook_vm.yml step in STIG metadata:
  - playbook: ../sssd-ci-containers/src/ansible/playbook_vm.yml extra_vars: override_stig_enabled: true

### DIFF
--- a/src/ansible/roles/packages/tasks/CentOS10.yml
+++ b/src/ansible/roles/packages/tasks/CentOS10.yml
@@ -50,6 +50,12 @@
     dnf:
       name: https://dl.fedoraproject.org/pub/epel/epel-release-latest-10.noarch.rpm
       state: present
+  # Tolerate failure on STIG-hardened systems where packages may be unsigned
+  rescue:
+  - name: Fail EPEL setup if not STIG
+    fail:
+      msg: "EPEL setup for passkey failed and stig_enabled is not set"
+    when: not stig_enabled | default(false) | bool
   when:
     - passkey_support
     - "'base_client' in group_names or 'client' in group_names"
@@ -65,6 +71,11 @@
     - systemd-devel
     - libtool
     - python3-koji
+  register: usbip_deps
+  # Tolerate failure on STIG-hardened systems where packages may be unsigned
+  failed_when:
+    - usbip_deps is failed
+    - not stig_enabled | default(false) | bool
   when:
     - passkey_support
     - "'base_client' in group_names or 'client' in group_names"
--- a/src/ansible/roles/packages/tasks/Fedora.yml
+++ b/src/ansible/roles/packages/tasks/Fedora.yml
@@ -294,6 +294,11 @@
       - go
       - nmap-ncat
       - git
+    register: passkey_packages
+    # Tolerate failure on STIG-hardened systems where packages may be unsigned
+    failed_when:
+      - passkey_packages is failed
+      - not stig_enabled | default(false) | bool
     when: passkey_support
   when: "'base_client' in group_names or 'client' in group_names or 'base_ipa' in group_names or 'ipa' in group_names"
 
@@ -302,6 +307,11 @@
     state: present
     name:
     - usbip
+  register: usbip_install
+  # Tolerate failure on STIG-hardened systems where packages may be unsigned
+  failed_when:
+    - usbip_install is failed
+    - not stig_enabled | default(false) | bool
   when:
     - passkey_support
     - "'base_client' in group_names or 'client' in group_names"
--- a/src/ansible/roles/packages/tasks/RedHat10.yml
+++ b/src/ansible/roles/packages/tasks/RedHat10.yml
@@ -10,6 +10,11 @@
     state: present
     name: 'https://dl.fedoraproject.org/pub/epel/epel-release-latest-{{ ansible_facts["distribution_major_version"] }}.noarch.rpm'
     disable_gpg_check: yes
+  register: epel_passkey
+  # Tolerate failure on STIG-hardened systems where packages may be unsigned
+  failed_when:
+    - epel_passkey is failed
+    - not stig_enabled | default(false) | bool
   when:
     - passkey_support
     - "'base_client' in group_names or 'client' in group_names"
@@ -25,6 +30,11 @@
     - systemd-devel
     - libtool
     - python3-koji
+  register: usbip_deps
+  # Tolerate failure on STIG-hardened systems where packages may be unsigned
+  failed_when:
+    - usbip_deps is failed
+    - not stig_enabled | default(false) | bool
   when:
     - passkey_support
     - "'base_client' in group_names or 'client' in group_names"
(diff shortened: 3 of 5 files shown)


madhuriupadhye:
This is extra var we need to add in downstream metadata,
```
      - playbook: ../sssd-ci-containers/src/ansible/playbook_vm.yml
        extra_vars:
            override_stig_enabled: true
```

src/ansible/roles/packages/tasks/CentOS10.yml
```diff
@@ -50,21 +50,33 @@
...
+      - automake
+      - autoconf
+      - systemd-devel
+      - libtool
+      - python3-koji
+  rescue:
+  - name: Fail usbip deps install if not STIG
+    fail:
+      msg: "usbip build dependencies installation failed and stig_enabled is not set"
+    when: not stig_enabled | default(false) | bool
```
spoore1: From looking into the block-rescue directive, it looks like it's more appropriate for blocks of multiple tasks.  I think a `failed_when` directive on the dnf task is enough for covering this.   I think it would just be:
```
- name: Install packages for usbip needed for virtual passkey testing
  dnf:
    state: present
    name:
    - gcc
    - make
    - automake
    - autoconf
    - systemd-devel
    - libtool
    - python3-koji
  when:
    - passkey_support
    - "'base_client' in group_names or 'client' in group_names"
  register: install_usbip_packages
  failed_when:
    - install_usbip_packages.rc != 0
    - not stig_enabled
```

Thinking about this some more, do we need to just skip the step if `stig_enabled` is True?   That might be even simpler:   just add "not stig_enabled" to the when list.

src/ansible/roles/packages/tasks/Fedora.yml
```diff
@@ -294,13 +294,25 @@
...
+  - name: Install usbip package
+    dnf:
+      state: present
+      name:
+      - usbip
+  rescue:
+  - name: Fail usbip install if not STIG
+    fail:
+      msg: "usbip installation failed and stig_enabled is not set"
+    when: not stig_enabled | default(false) | bool
```
spoore1: Same comment as the other dnf task.  I think a `failed_when` directive for the dnf task would be enough here.  Or adding a "when: not stig_enabled" for the dnf task if we should just be skipping it entirely for STIG.

src/ansible/roles/packages/tasks/RedHat10.yml
```diff
@@ -6,25 +6,39 @@
...
+  - name: Install EPEL for passkey
+    dnf:
+      state: present
+      name: 'https://dl.fedoraproject.org/pub/epel/epel-release-latest-{{ ansible_facts["distribution_major_version"] }}.noarch.rpm'
+      disable_gpg_check: yes
+  rescue:
+  - name: Fail EPEL install if not STIG
+    fail:
+      msg: "EPEL installation for passkey failed and stig_enabled is not set"
+    when: not stig_enabled | default(false) | bool
```
spoore1: Same `failed_when` comment here.

src/ansible/roles/packages/tasks/RedHat10.yml
```diff
@@ -6,25 +6,39 @@
...
+      - automake
+      - autoconf
+      - systemd-devel
+      - libtool
+      - python3-koji
+  rescue:
+  - name: Fail usbip deps install if not STIG
+    fail:
+      msg: "usbip build dependencies installation failed and stig_enabled is not set"
+    when: not stig_enabled | default(false) | bool
```
spoore1: Same `failed_when` comment here as well.

src/ansible/roles/passkey/tasks/main.yml
```diff
@@ -74,4 +79,9 @@
     command: modprobe vhci_hcd
     ignore_errors: true
 
+  rescue:
+  - name: Fail vfido setup if not STIG
+    fail:
+      msg: "Vfido setup failed and stig_enabled is not set"
+    when: not stig_enabled | default(false) | bool
```
spoore1, line 88: This is probably a good use of block-rescue since it encompasses multiple tasks.

src/ansible/roles/passkey/tasks/main.yml
```diff
@@ -17,6 +17,11 @@
     file:
       path: /opt/random.c
       state: absent
+  rescue:
+  - name: Fail passkey build if not STIG
+    fail:
+      msg: "Passkey build failed and stig_enabled is not set"
+    when: not stig_enabled | default(false) | bool
```
spoore1, line 25: Good use of block-rescue from what I read about it being useful for multiple tasks.

src/ansible/roles/packages/tasks/Fedora.yml
```diff
@@ -294,13 +294,25 @@
       - nmap-ncat
       - git
     when: passkey_support
+  rescue:
+  - name: Fail passkey package install if not STIG
+    fail:
+      msg: "Passkey package installation failed and stig_enabled is not set"
+    when: not stig_enabled | default(false) | bool
   when: "'base_client' in group_names or 'client' in group_names or 'base_ipa' in group_names or 'ipa' in group_names"
```
spoore1: I'm kind of torn on this one since it's already a block.   However, to be consistent with the rest of my suggestions for dnf tasks, I think it can probably be handled the same as those with `failed_when` instead of a block-rescue.

src/ansible/roles/packages/tasks/CentOS10.yml
```diff
@@ -50,21 +50,33 @@
     dnf:
       name: https://dl.fedoraproject.org/pub/epel/epel-release-latest-10.noarch.rpm
       state: present
+  rescue:
+  - name: Fail EPEL setup if not STIG
+    fail:
+      msg: "EPEL setup for passkey failed and stig_enabled is not set"
+    when: not stig_enabled | default(false) | bool
```
spoore1, line 58: I think this can be a `failed_when` directive like the others I mention later.   block-rescue from what I found is more appropriate for multiple tasks than single ones.

spoore1 (changes requested):
I think most of the `block-rescue` implementations here (especially for the single dnf tasks) would be better as simple `failed_when` directives added to the task.   This would simplify the task and overall readability of the tasks files.   Most of the cases are single task cases which from what I read are better handled with `failed_when`.   There are a few multiple task blocks at the end that look like a good use of `block-rescue` so I commented as such there.   

If my findings about `block-rescue` are not accurate, please let me know and I will review again with that in mind.

src/ansible/roles/packages/tasks/CentOS10.yml
spoore1: Since it's already a block, I'll drop this comment.

spoore1 (approved):
Updates look good to me.   Just double check that the changes work with the STIG tests in CI and we can merge this one I think.   Otherwise, I can run tests next week to make sure it's all good.

spoore1:
Failures appear to be due to github infrastructure issues.  I ran a build locally with this change with no issues.   Merging this one now.

PR #205, opened 2026-08-19
actions: add timeout and retry logic when post-processing containers

pbrezina:
For some reason, `podman exec` started to hang in Github Actions
environment - the hang was mostly when touching the ldap containers, but
I also saw it with nfs. I do not know what is the problem, probably not
on our side.

This retry logic seems to help, hopefully it fix it.


pbrezina:
Retry logic can get over the setup place, but other parts that talks to the container are affected as well and would have to be retried... so this is not the way.

PR #211, opened 2026-08-26
ci: Split installing ansible for Fedora 45 and newer

jakub-vavra-cz:
Bootstrap python 3.13 and ansible from github devel branch for Fedora 45 and 46 builds.
Disable h264 repo if present because it has broken gpg keys on rawhide (f46) now.
Workaround overlay issue on C8S in virtsmartcard.


thalman:
I wonder if we can simply use `runs-on: ubuntu-26.04` (or ubuntu-22.04) instead this

jakub-vavra-cz:
> I wonder if we can simply use `runs-on: ubuntu-26.04` (or ubuntu-22.04) instead this

I expect that at some point the 26.04 will become the ubuntu-latest.
For now lets stay with the current one as it does not have the freezing podman.
I hope that in october the official ansible 2.22 is released and we would be able to drop part of this mess.
Unfortunately the 2.22 is dropping support of python 3.12 on controller officially so it is not just devel branch.

src/ansible/roles/packages/tasks/Fedora.yml
```diff
@@ -342,8 +342,15 @@
...
+  - name: Check if fedora-cisco-openh264 repo exists
+    shell: dnf repolist all fedora-cisco-openh264
+    ignore_errors: true
+    failed_when: false
+    changed_when: false
+    register: openh264_repo
+
   - name: Install SSSD build dependencies
-    command: dnf builddep -y sssd
+    command: "dnf builddep -y{% if openh264_repo.rc == 0 %} --disablerepo=fedora-cisco-openh264{% endif %} sssd"
```
alexey-tikhonov:
```
  The check dnf repolist all fedora-cisco-openh264 exits 0 whether or not the repo exists
(repolist only filters output; all is a bogus positional here), so openh264_repo.rc == 0 is
 effectively always true and   --disablerepo=fedora-cisco-openh264 gets appended unconditionally.
On an image lacking that repo, if dnf ever errors on disabling a nonexistent repo, dnf builddep
sssd breaks. The working idiom already lives a few lines down (line 149, beaker-tasks): pipe to 
grep so rc reflects the match — yum repolist | grep fedora-cisco-openh264.
```

src/ansible/roles/virtsmartcard/tasks/main.yml
```diff
@@ -63,6 +63,26 @@
...
+    store=$parent/targeted
+    if [ ! -d "$store" ]; then
+      exit 0
+    fi
+    rm -rf "$parent/targeted.overlay-fix"
+    cp -a "$store" "$parent/targeted.overlay-fix"
+    rm -rf "$store"
+    mv "$parent/targeted.overlay-fix" "$store"
+  args:
+    creates: "{{ virt_smartcard_dir }}/virtcacard.cil.done"
```
alexey-tikhonov: """
  The overlay-copy task uses creates: .../virtcacard.cil.done, but that marker is created by the later "Install custom semodule" task (line 91), not this one. This couples the guard to an unrelated side effect:
  renaming/reordering the semodule marker silently disables the overlay fix, reintroducing the EXDEV/ENOTEMPTY rename failure on C8S with no obvious cause.
"""

.github/actions/install-dependencies/action.yml
```diff
@@ -10,23 +10,43 @@ inputs:
...
+    if: ${{ inputs.tag != 'centos-8' && inputs.tag != 'fedora-45' && inputs.tag != 'fedora-46' }}
     shell: bash
     run: |
       sudo pip3 install ansible passlib
 
-  - name: Install deb packages
+  # Install ansible from devel branch for Fedora 45 and 46 for python 3.15
+  # This ugly hack is needed because the latest ansible version available
+  # in the official repositories is too old and unreleased 2.22 is needed.
+  # Unfortunately current devel ansible needs python 3.13...
```
alexey-tikhonov: I'm confused: first line says "for python 3.15" and this line says "needs python 3.13" (that is actually installed)

.github/actions/install-dependencies/action.yml
```diff
@@ -10,23 +10,43 @@ inputs:
...
+  - name: Install ansible from devel branch
+    if: ${{ inputs.tag == 'fedora-45' || inputs.tag == 'fedora-46' }}
     shell: bash
     run: |
-      sudo apt-get update
-      sudo apt-get install -y docker-compose
+      sudo apt install software-properties-common gpg -y
+      sudo add-apt-repository ppa:deadsnakes/ppa -y
+      sudo apt update
+      sudo apt install python3.13 python3.13-venv python3.13-dev -y
```
alexey-tikhonov: It should use 'apt-get', as above.

src/ansible/roles/packages/tasks/Fedora.yml
```diff
@@ -342,8 +342,15 @@
 
 - name: Install additional packages for client development image
   block:
+  - name: Check if fedora-cisco-openh264 repo exists
+    shell: dnf repolist all fedora-cisco-openh264
+    ignore_errors: true
+    failed_when: false
```
alexey-tikhonov: Isn't using both 'failed_when' and 'ignore_errors' redundant?

src/ansible/roles/packages/tasks/Fedora.yml
```diff
@@ -342,8 +342,15 @@
 
 - name: Install additional packages for client development image
   block:
+  - name: Check if fedora-cisco-openh264 repo exists
```
thalman: This commit has really poor message.

Despite that can't we disable that repo for good? I think that SSSD does not need it at all.

thalman (changes requested):

src/ansible/roles/packages/tasks/Fedora.yml
thalman: yes, in this case `ignore_errors` is redundant.

.github/actions/install-dependencies/action.yml
jakub-vavra-cz: For managed system with python 3.15 we need ansible 2.22 on controller (ubuntu) that needs python 3.13 on while the ubuntu-latest (24.04) contains only 3.12.

.github/actions/install-dependencies/action.yml
jakub-vavra-cz: It does not matter as I am not parsing its output.  The choice is equal if go for apt-get that should be stable or apt that is focused on user friendly output.

alexey-tikhonov: At the very least it's inconsistent.

src/ansible/roles/packages/tasks/Fedora.yml
jakub-vavra-cz: fixed

src/ansible/roles/virtsmartcard/tasks/main.yml
jakub-vavra-cz: updated

.github/actions/install-dependencies/action.yml
jakub-vavra-cz: switched to apt-get

src/ansible/roles/packages/tasks/Fedora.yml
jakub-vavra-cz: done

src/ansible/roles/packages/tasks/Fedora.yml
jakub-vavra-cz: changed

src/ansible/roles/packages/tasks/Fedora.yml
```diff
@@ -16,6 +16,26 @@
   set_fact:
     extended_packageset: "{{ extended_packageset | default('no') }}"
 
+# Broken GPG keys / missing source metalink on Fedora rawhide (F46).
+# Disable by repo id (dnf5 override) so this does not depend on the
+# .repo path. SSSD does not need cisco-openh264.
+- name: Disable fedora-cisco-openh264 repositories
+  shell: |
+    if command -v dnf5 >/dev/null; then
```
thalman: can we use https://docs.ansible.com/projects/ansible/latest/collections/community/general/dnf_config_manager_module.html instead of scripting it?

src/ansible/roles/packages/tasks/Fedora.yml
```diff
@@ -343,7 +363,7 @@
 - name: Install additional packages for client development image
   block:
   - name: Install SSSD build dependencies
-    command: dnf builddep -y sssd
+    command: dnf builddep -y --disablerepo=fedora-cisco-openh264* sssd
```
thalman: This change can be reverted now

thalman (changes requested):

.github/actions/install-dependencies/action.yml
alexey-tikhonov: Thank you for the update.

.github/actions/install-dependencies/action.yml
alexey-tikhonov: Thank you.

src/ansible/roles/packages/tasks/Fedora.yml
alexey-tikhonov: I agree and I don't have other additional comments besides this.

jakub-vavra-cz: done

src/ansible/roles/packages/tasks/Fedora.yml
jakub-vavra-cz: Switched just to delete file instead. The last thing I want to is add additional dependency on collection. Bootstrapping dev ansible in additional python is already a pain even without involving more collections.

thalman: understood, if dnf_config_manager is not in default setup then it make sense.

I would personally prefer to disable it like this (just to have the installation consistent):

```
- name: Disable cisco openh264 repo
  ansible.builtin.replace:
    path: /etc/yum.repos.d/fedora-cisco-openh264.repo
    regexp: '^enabled\s*=.*$'
    replace: 'enabled = 0'
```

but I do not insist on that.

thalman (approved):

src/ansible/roles/packages/tasks/Fedora.yml
jakub-vavra-cz: Keeping the delete as I burned already too much time on this, to redo it again.

alexey-tikhonov (approved):

alexey-tikhonov:
Let's wait for CI to complete.

alexey-tikhonov:
f-46:
```
[7](https://github.com/SSSD/sssd-ci-containers/actions/runs/33366436004/job/99407990005?pr=211#step:4:10489)
          Importing OpenPGP key 0x91211FCE:
           UserID     : "Fedora (46) <fedora-46-primary@fedoraproject.org>"
           Fingerprint: D924B10D3E810DABDD8B56B596E7E91491211FCE
           From       : file:///usr/share/pki/rpm-gpg/RPM-GPG-KEY-fedora-rawhide-x86_64
          The key was successfully imported.
          Transaction failed: Signature verification failed.
          OpenPGP check for package "openh264-2.6.0-3.fc45.x86_64" (/var/cache/libdnf5/fedora-cisco-openh264-4896e02bbb10d47b/packages/openh264-2.6.0-3.fc45.x86_64.rpm) from repo "fedora-cisco-openh264" has failed: Import of the key didn't help, wrong key?
```

jakub-vavra-cz:
> f-46:
> 
> ```
> [7](https://github.com/SSSD/sssd-ci-containers/actions/runs/33366436004/job/99407990005?pr=211#step:4:10489)
>           Importing OpenPGP key 0x91211FCE:
>            UserID     : "Fedora (46) <fedora-46-primary@fedoraproject.org>"
>            Fingerprint: D924B10D3E810DABDD8B56B596E7E91491211FCE
>            From       : file:///usr/share/pki/rpm-gpg/RPM-GPG-KEY-fedora-rawhide-x86_64
>           The key was successfully imported.
>           Transaction failed: Signature verification failed.
>           OpenPGP check for package "openh264-2.6.0-3.fc45.x86_64" (/var/cache/libdnf5/fedora-cisco-openh264-4896e02bbb10d47b/packages/openh264-2.6.0-3.fc45.x86_64.rpm) from repo "fedora-cisco-openh264" has failed: Import of the key didn't help, wrong key?
> ```

They moved repos to /usr/share/dnf5/repos.d/, I had to update it again :facepalm:

PR #213, opened 2026-08-31
ci: Lock Centos-8 build with Fedora-44 tag for samba.

jakub-vavra-cz:
Ansible that can manage Centos-8 can manage Fedora only upto version 44 due to incompatible python.

install-dependencies/action.yml
1. Everything that is not c8s uses ansible devel until 2.22 is released.
2. c8s uses locked old version of ansible 9.8.

build.sh
When samba with the same tag (centos-8, centos-9, centos-10) is unavailable:
1. Use samba on fedora-44 with centos-8.
2. Otherwise use samba on latest (= fedora 45).


src/build.sh
```diff
@@ -92,9 +92,18 @@ function build_base_image {
     # base images: UNAVAILABLE=samba matches name=base-samba -> ci-base-samba
     # other images: UNAVAILABLE=client-devel matches name=client-devel -> ci-client-devel
     if [ "base-$svc" = "$name" ]; then
-      quay_image="quay.io/sssd/ci-base-$svc:latest"
+      # Ansible working with centos-8 supports Fedora upto 44.
+      if [ "$TAG" == "centos-8" && "$svc" == "samba" ]; then
```
alexey-tikhonov: Is this valid sytnax?
Shouldn't it be `[ ... ] && [ ... ]` or `[[ ... && ... ]]`?
Here and below.

.github/actions/install-dependencies/action.yml
```diff
@@ -22,18 +22,19 @@ runs:
...
-    run: |
-      sudo pip3 install ansible passlib
+  ## TODO: Uncomment this when ansible-core 2.22 is available in the official repositories
+  #- name: Install python packages
+  #  if: ${{ inputs.tag != 'centos-8' }}
+  #  shell: bash
+  #  run: |
+  #    sudo pip3 install ansible passlib
 
   # Install ansible from devel branch to manage Fedora 45 and 46 with python 3.15
```
alexey-tikhonov: If it's really required to install ansible from devel everywhere besides c8s, then this comment needs to be updated.

src/build.sh
jakub-vavra-cz: You are right. Fixed.

.github/actions/install-dependencies/action.yml
jakub-vavra-cz: Updated the comments.

alexey-tikhonov: Why does new comment say "Fedora >= 45" and not "Fedora >= 44" or "everywhere besides c8s"?

jakub-vavra-cz: "Install ansible from devel branch to manage Fedora >= 45 with python 3.15"

Because the the root cause that we need devel is Fedora 45+. 
Fedora 44 is works with the old one without and issue.

alexey-tikhonov (approved):

thalman (approved):
ACK

PR #214, opened 2026-09-01
setup: create wrapper for docker-compose

pbrezina:
The docker-compose command is not directly available in ubuntu-26-04
image. Creating a wrapper around it to avoid changing scripts for now.

### DIFF
--- a/actions/setup/action.yml
+++ b/actions/setup/action.yml
@@ -28,6 +28,15 @@ runs:
       sudo apt-get update
       sudo apt-get install -y docker-compose
 
+      # Create wrapper script for docker-compose that uses podman compose
+      # /usr/bin/docker-compose is not available on ubuntu-26-04 even when
+      # installed
+      if ! command -v docker-compose &> /dev/null; then
+        echo -e '#!/bin/bash\nexec podman compose "$@"' | sudo tee /usr/bin/docker-compose > /dev/null
+        sudo chmod +x /usr/bin/docker-compose
+        docker-compose --help
+      fi
+
   - name: Start podman socket
     shell: bash
     run: |


actions/setup/action.yml
```diff
@@ -28,6 +28,15 @@ runs:
       sudo apt-get update
       sudo apt-get install -y docker-compose
 
+      # Create wrapper script for docker-compose that uses podman compose
+      # /usr/bin/docker-compose is not available on ubuntu-26-04 even when
+      # installed
+      if ! command -v docker-compose &> /dev/null; then
+        echo -e '#!/bin/bash\nexec podman compose "$@"' | sudo tee /usr/bin/docker-compose > /dev/null
+        sudo chmod +x /usr/bin/docker-compose
+        docker-compose --help
```
alexey-tikhonov, line 37: What is the point of `docker-compose --help`?

actions/setup/action.yml
```diff
@@ -28,6 +28,15 @@ runs:
       sudo apt-get update
       sudo apt-get install -y docker-compose
```
alexey-tikhonov, line 29: Is it ignored on Ubuntu 26?

actions/setup/action.yml
pbrezina: to make sure the command works as expected

alexey-tikhonov: Then process exit code?
Or is it for verbosity in the logs only?

actions/setup/action.yml
pbrezina: technically, it is not needed for either version because it is already preinstalled. It is now a docker plugin "docker compose" rather then standalone binary.

pbrezina: I think @justin-stephenson had already patches that converts from docker-compose to docker compose all over the place, he can resurrect it and then we can drop this completely. But let's fix PR CI first.

actions/setup/action.yml
pbrezina: `set -ex` -> if it fails, the script fails

alexey-tikhonov (approved):

sumit-bose (approved):
Hi,

thank you for the fix, ACK.

bye,
Sumit

PR #216, opened 2026-09-08
ad: Changes for enabling AD forest topology

jakub-vavra-cz:
Update dns for ad in idmci.
Update ad role to work with ad forest: The sudo schema is now imported on ad tree.
Update ipasmartcard not to hard crash when ipa server is not part of the topology.
Make sure that certificate templates are installed only on the top ad.
Tweak choco install to be more resilient.


thalman (approved):
LGTM

danlavu (approved):

PR #217, opened 2026-09-09
Ensure .Net 4.8 for Win2019

spoore1:
Chocolatey version 2.0.0 requires .Net 4.8.   Adding .Net update to the
"ad" role to make sure it gets installed for Windows 2019 servers.


madhuriupadhye (approved):
lgtm

